PR #57, opened 2023-02-07
Refactored, added tests and docs to layers table class

DrPaulSharp:
Added tests for the renamed class "layersClassRealSLD.m"
Refactored both "layersClassRealSLD.m" and calling routines in "projectClass.m"


API/projectClass/layersClassRealSLD.m
```diff
@@ -0,0 +1,293 @@
...
+        
+        function obj = setLayerValue(obj, inputValues, paramNames)
+            % Change the value of a given layer parameter in the table
+            % (excluding the layer name). The row and column of the
+            % parameter can both be specified by either name or index.
+            % The expected input is a cell array consisting of three
+            % values: row, column, newValue, and a string array of
+            % parameter names defined in the project's parameter class.
+            %
+            % layers.setLayerValue(1, 1, "origin", parameters.paramsTable{:, 1});
```
StephenNneji, line 108: This example doesn't match the actual usage `layers.setLayerValue({1, 1, "origin"}, parameters.paramsTable{:, 1});`. I prefer the syntax in the docstring, you can use a varargin then take varargin{end} as the table then varargin{1:end-1} as the inputs. Another option would be to flip the argument order, you could do` function obj = setLayerValue(obj, paramNames, varargin)` which should allow you do `layers.setLayerValue(parameters.paramsTable{:, 1}, 1, 1, "origin");`

API/projectClass/layersClassRealSLD.m
```diff
@@ -0,0 +1,293 @@
...
+            varTypes = {'string','string','string','string','string','string'};
+            varNames = {'Name','Thickness','SLD','Roughness','Hydration','Hydrate with'};
+            obj.layersTable = table('Size',sz,'VariableTypes',varTypes,'VariableNames',varNames);
+            %obj.layersTable(1,:) = {'Layer 1','', '', '', '','bulk out'};
+            obj.layersCount = 0;
+            obj.layersAutoNameCounter = 1;
+            
+        end
+        
+        function obj = addLayer(obj, whatToAdd, paramNames)
```
StephenNneji, line 31: See previous comment, could be better to flip the argument order and use a varargin for whatToAdd

API/projectClass/layersClassRealSLD.m
```diff
@@ -0,0 +1,293 @@
+classdef layersClassRealSLD < handle
+    
+    % This is the class definition for the layers block with no absorption.
+
+    properties
+        layersTable = table
+    end
+    
+    properties (Access = private)
+        layersCount
```
StephenNneji, line 10: recommend
```
   properties (Dependent)
        layersCount;
    end
    
      function count = get.layersCount(obj)
           count = height(obj.layersTable);
      end
```

API/projectClass/layersClassRealSLD.m
```diff
@@ -0,0 +1,293 @@
...
+            % (either name or index) and a list of parameter names.
+
+            if ischar(inputVal)
+                if ~any(strcmpi(inputVal, paramNames))
+                    error('Parameter %s not recognized', inputVal);
+                end
+                param = inputVal;
+
+            elseif isnumeric(inputVal)
+                paramIndex = floor(inputVal);
```
StephenNneji, line 276: I missed this in the multitype table: I don't think it is a good idea to mask a bad index like this. A user indexing with 5.5 could mean to index with 5 or 6 but this assumes we know what the user wants. Also the user might be calling the wrong function and will not know because we correct their input. In my opinion, it is a well defined behaviour for a non-integer index to crash

StephenNneji (approved):
Please address comments

API/projectClass/layersClassRealSLD.m
```diff
@@ -0,0 +1,293 @@
...
+            
+        end
+
+        function removeLayer(obj, layer)
+            % Removes a layer from the layers table. The expected input is
+            % a length one cell array.
+            % NOTE that an input such as {[1 3]} leads to multiple rows
+            % being removed from the table
+            %
+            % layers.removeLayer({2})
```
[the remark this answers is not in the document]
DrPaulSharp: It was to do with the call from project class, which has now been changed to resolve this.

API/projectClass/layersClassRealSLD.m
DrPaulSharp: The input has been changed in both the tests and the project class to match that in the docstring.

API/projectClass/layersClassRealSLD.m
DrPaulSharp: Done, and allows removal of some now unnecessary code from projectClass

PR #59, opened 2023-02-10
Added tests and docs to custom file class

DrPaulSharp:
Adds the class "testCustomFileClass" which sets out a suite of explicit tests for the custom file class within project class.
Adds custom RAT exceptions to utilities/exceptions and implements them in "customFileClass.m"
Tests for custom exceptions included in "testExceptions"


utilities/exceptions/invalidInput.m
```diff
@@ -0,0 +1,10 @@
+classdef invalidInput < MException
```
StephenNneji, line 1: What the difference btw invalid type and invalid input?

tests/testMultiTypeTable.m
```diff
@@ -48,13 +48,13 @@
...
 
     properties
-        exampleTable;           % Example Multi-Type Table for testing
-        initialTypesTable;
-        initialAllowedTypes = {'constant'  'data'  'function', 'gaussian'};
-        initialAllowedActions = {'add'  'subtract'};
-        initialTypesCount = 1;
-        initialTypesAutoNameCounter = 1;
-        initialTypesAutoNameString = 'Row';
+        exampleTable            % Example Multi-Type Table for testing
```
StephenNneji, line 51: I don't think there is any benefit to semicolon so we  have to agree a style an unify all

utilities/exceptions/invalidInput.m
DrPaulSharp: As I was developing the exceptions Invalid input split into invalid type and invalid number of inputs. I'll remove this one and we can always readd it later if needs be.

tests/testMultiTypeTable.m
DrPaulSharp: I'd spotted this as well. I'll take the semicolons out, which I prefer.

StephenNneji (approved):

PR #80, opened 2023-02-16
Fixes inconsistent model type

StephenNneji:
Fixes uses of "Layers" as the model type. 
Removes offending mex file and updates test data


testRatScript.m
```diff
@@ -23,8 +23,6 @@
 disp('Adding paths Initiated')
 addRatPaths;
 disp('Adding paths Completed')
-disp('Compiling RAT to mex Initiated')
```
StephenNneji, line 26: Apparently the tests never compile the mex

DrPaulSharp (commented):
Given that the domains classes are still included on the master branch I think it would be worth making these changes there also. There are five instances of 'layers' in domains related routines:

API/domainsClass/domainsClass.m: line 227 & line 277
API/domainsClass/reflContrastsClass.m: line 250
API/domainsClass/simContrastsClass.m: line 156
targetFunctions/domains_TF/domainsTF_reflectivityCalculation.m: line 66

RabiyaF (approved):

PR #86, opened 2023-02-22
Adds test for the resolution class and updates docs

StephenNneji:


tests/testResolutionsClass.m
```diff
@@ -0,0 +1,283 @@
...
+            testCase.verifyError(@setShowPriors, invalidType.errorID);  % showPrior should be logical 
+            function setShowPriors
+                resolution.showPriors = 'a';
+            end
+            resolution.showPriors = true;
+        end
+        
+        function testGetResolNames(testCase)
+            % Tests getResolNames returns correctly  
+            resolution = resolutionsClass(testCase.parameters(1, :), testCase.resolutions(1, :));
```
DrPaulSharp, line 54: It would be worth defining a resolutions class in the method setup, given that this line appears in several tests.

tests/testResolutionsClass.m
```diff
@@ -0,0 +1,283 @@
...
+            displayArray = textscan(display{2},'%s','Delimiter','\r','TextType','string');
+            displayArray = strip(displayArray{1});
+            % display table should be height of resolPars table plus header and divider row
+            testCase.verifyLength(displayArray, height(resolution.resolPars.paramsTable) + 3);
+            % Remove html tags used to format header then split table when
+            % 2 or more spaces are found to avoid splitting names with single space
+            displayHeader = eraseBetween(displayArray{2}, '<', '>', 'Boundaries','inclusive');
+            displayHeader = regexp(displayHeader, '\s{2,}', 'split');
+            testCase.verifyEqual(displayHeader, paramHeader);
+	        row = regexp(displayArray{4}, '\s{2,}', 'split');
```
DrPaulSharp, line 217: Why is this line (and others below) indented?

tests/testResolutionsClass.m
```diff
@@ -0,0 +1,283 @@
+classdef testResolutionsClass < matlab.unittest.TestCase   
+    
+    properties
+        parameters
+        resolutions
+    end
+
+    methods(TestMethodSetup)
+        function createResolutions(testCase)
```
DrPaulSharp, line 9: Given that these properties are not modified in any of the tests, the setup method should be at the class level rather than method level - i.e., "methods(TestClassSetup)" - there's no need to re-initialise the parameters and resolutions before each individual test. Of course, the resolutions class "resolution" would still need to be initialised at the method level.

DrPaulSharp (approved):

API/projectClass/resolutionsClass.m
```diff
@@ -195,29 +211,21 @@
             end
```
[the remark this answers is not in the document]
DrPaulSharp: Helpful . . . I'm referring to the index-based clauses in the if statement above.

tests/testResolutionsClass.m
StephenNneji: I have no idea, it doesn't show up in the MATLAB editor but shows up in vscode. Now fixed

PR #88, opened 2023-02-27
Adds tests and docs to the contrasts class

DrPaulSharp:


API/projectClass/contrastsClass.m
```diff
@@ -105,127 +124,148 @@
...
         function contrastNames = getAllContrastNames(obj)
+            % Get the names of all contrasts defined in the class.
+            %
+            % contrasts.getAllContrastNames()
             nContrasts = obj.numberOfContrasts;
             contrastNames = cell(1,nContrasts);
             
             if nContrasts == 0
-                error('Can''t get contrast names when no contrasts defined')
+                throw(invalidValue('Can''t get contrast names when no contrasts defined'));
```
StephenNneji, line 268: Do you know why this throws an error instead of just returning an empty array?

StephenNneji (approved):

API/projectClass/contrastsClass.m
```diff
@@ -1,20 +1,42 @@
...
+
         function obj = addContrast(obj, allowedNames, varargin)
+            % Add a contrast to the class
+            % A class can be added with no input parameters, just a class
+            % name, or a set of key-value pairs.
+            %
+            % contrasts.addContrast()
+            % contrasts.addContrast('New Contrast')
+            % contrasts.addContrast('name', 'new contrast', ... 
+            %                       'background', 'Background H2O')
```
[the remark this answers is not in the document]
DrPaulSharp: That was there before, I'll trim it.

API/projectClass/contrastsClass.m
DrPaulSharp: Good point, there doesn't appear to be any reason to me, as this routine is only used within the class and we check if the result is empty when we call it. I'll remove this clause.

PR #101, opened 2023-03-17
Changed strings to characters

### LINKED ISSUE #63: Inconsistent use of char array and string
There is an inconsistent use of strings and char array all over the codebase. This should be fixed by creating an `isText` function in utilities that ensures both char array and 1 length string arrays are  allowed then all variables should be casted into a string (or char array if its better for coder)

DrPaulSharp:
Fixes #63 

Converted strings to characters in the code, alongside using character conversion instead of string conversion where possible.
Note that strings are still widely used in tests, as they are the class used for table values.


utilities/paths.m
```diff
@@ -44,11 +44,11 @@ function paths()
 
         
     elseif ispc % WINDOWS
-        arch = string(computer('arch'));
-        %incPath1= matlabRoot+'\'+"extern"+'\'+"include";
-        incPath1= 'C:\Program Files\MATLAB\R2021a\extern\include'
+        arch = char(computer('arch'));
```
StephenNneji, line 47: `computer('arch')` returns a char array so conversion is not needed

utilities/paths.m
```diff
@@ -24,12 +24,12 @@ function paths()
 
 
     elseif isunix %LINUX
-        arch = string(computer('arch'));
-        incPath1= matlabRoot+'/'+"extern"+'/'+"include";
+        arch = char(computer('arch'));
```
StephenNneji, line 27: Similar to previous comment

StephenNneji (commented):
Thanks for tackling this. I think the user facing functions should accept text char/string then convert to char but internal method should expect char

API/projectClass/projectClass.m
```diff
@@ -777,7 +777,7 @@
             
             % parse the layers details
             layersValues = layersCell(:,2:end);
-            paramNames = string(paramStruct.paramNames);
+            paramNames = char(paramStruct.paramNames);
```
StephenNneji, line 780: I think `paramNames` is converted to char in `parametersClass.toStruct` ? so this might not be needed

utilities/isText.m
StephenNneji, line 1: Please add a small unit test for this in testUtilites

StephenNneji (approved):

API/projectClass/projectClass.m
StephenNneji: This changed has broken the repo. Sorry I did not mean removed the line I meant remove the char conversion

PR #108, opened 2023-03-24
fixes path clash for orsoDSPCCustomLayers

RabiyaF:
The example paths are now added when the testExamples class is SetUp.
The paths are then removed during class teardown.

The paths to the examples have been removed from  `addPaths.m`

A minor update is also made to `makeInputsAndOutputs.m` because the `.mat` files where not being generated in the right folders on Mac.


tests/testExamples.m
```diff
@@ -1,61 +1,89 @@
...
+    end
+
+    methods(TestClassTeardown)
         function testCleanUp(testCase)
+            root = getappdata(0,'root');
+            for i = 1:length(testCase.exampleFolder)
+                rmpath(fullfile(root, 'examples', testCase.exampleFolder{1,i}));
+            end
+            % Generated by the DSPCMonolayerTwoContrasts example
             if exist('twoContrastExample.mat', 'file') == 2
```
StephenNneji, line 87: This deletes all instances of twoConstrastExample.mat from the repo including the ones in the examples folder. Please remove this and replace with 

```
    methods (TestClassSetup)
        function setWorkingFolder(testCase)
            % Makes a temporary folder the current working directory so 
            % file written by the example are deleted 
            import matlab.unittest.fixtures.WorkingFolderFixture;
            
            testCase.applyFixture(WorkingFolderFixture);
        end
    end
```

tests/testExamples.m
StephenNneji, line 1: Please delete the following:

- tests/twoContrastExample.mat
- examples/manual/DSPCCustomLayers/twoContrastExample.mat

tests/testExamples.m
```diff
@@ -1,64 +1,96 @@
...
+    methods(TestClassSetup)
+        function addExamplePaths(testCase)
+            root = getappdata(0,'root');
+            for i = 1:length(testCase.exampleFolder)
+                addpath(fullfile(root, 'examples', testCase.exampleFolder{1,i}));
+            end
+        end
+    end
+
+    methods (TestClassSetup)
```
StephenNneji, line 80: You don't have to duplicate the TestClassSetup block just put the setWorkingFolder function in the same block as the addExamplePaths function

examples/manual/DSPCStandardLayers/twoContrastExample.mat
StephenNneji, line 1: please restore

tests/testExamples.m
RabiyaF: Consolidated the two setUp methods

tests/testExamples.m
```diff
@@ -1,64 +1,89 @@
...
             4,...
             6,...
-            3};
+            3,...
+            4};
 
     end
 
-    methods (TestMethodTeardown)
+    methods(TestClassSetup)
```
StephenNneji, line 71: last change, I think this is better
```
    methods (TestClassSetup)
        function addExamplePaths(testCase)
            root = getappdata(0,'root');
            for i = 1:length(testCase.exampleFolder)
                addpath(fullfile(root, 'examples', testCase.exampleFolder{1,i}));
            end
        end

        function setWorkingFolder(testCase)
            % Makes a temporary folder the current working directory so 
            % file written by the example are deleted 
            import matlab.unittest.fixtures.WorkingFolderFixture;
            
            testCase.applyFixture(WorkingFolderFixture);
        end
    end
```

StephenNneji (approved):

PR #122, opened 2023-04-24
Removes unused files and fixes testExamples not running

StephenNneji:


StephenNneji:
@arwelHughes I removed utilities/paths.m, is this used for anything?

PR #127, opened 2023-04-26
Adds an example using the different minimizers

StephenNneji:
Removes examples/originalDSPCBilayer 
Renames 'originalDSPCBilayerStructInput.mat' and 'originalDSPCBilayerProjectClass.mat' to remove 'original' prefix
Adds controlAlgorithms.m to example


tests/testExamples.m
```diff
@@ -11,66 +12,88 @@
             'DSPCMonolayerStandardLayers',...
             'DSPCMonolayerTwoContrasts',...
             'orsoDSPCCustomLayers',...
-            'projectConversion'};
+            'projectConversion',...
+            'DSPCMonolayerCustomLayers',...
+            'controlAlgorithms'};
```
DrPaulSharp, line 17: The "controlAlgorithms" test adds considerably to the runtime and processing requirements for the test suite. Does it need to be run in full every time?

examples/controlAlgorithms.m
```diff
@@ -11,32 +11,32 @@
...
+% plotRefSLD(problem2, result2);
+% 
+% figure(3); clf
+% plotRefSLD(problem3, result3);
+% 
+% figure(4); clf
+% plotRefSLD(problem4, result4);
+% 
+% figure(5); clf
+% plotRefSLD(problem5, result5);
```
DrPaulSharp, line 42: Am I right in thinking that RAT is run twice here, but only one set of results is plotted?

DrPaulSharp (approved):
Might be worth thinking about how many iterations of simplex are required to satisfy what needs to be tested here - there's no point in spinning the algorithm's wheels.

examples/controlAlgorithms.m
StephenNneji: Missed that line

StephenNneji:
> Might be worth thinking about how many iterations of simplex are required to satisfy what needs to be tested here - there's no point in spinning the algorithm's wheels.

Since it converges after a single iteration, I can set this to 1 maybe

DrPaulSharp:
> > Might be worth thinking about how many iterations of simplex are required to satisfy what needs to be tested here - there's no point in spinning the algorithm's wheels.
> 
> Since it converges after a single iteration, I can set this to 1 maybe

Yeah, that's what I was thinking

PR #131, opened 2023-05-10
Updates docs to reflect  API changes and moves userManual to tutorial

StephenNneji:
- Updates docs to reflect  API changes
- Moves userManual.rst to tutorial.rst


docs/source/chapter2.rst
```diff
@@ -1,14 +1,8 @@
...
-...................................
-
-Basics
-======
 
+The Problem Definition Input Class
+..................................
 In the previous chapter, we saw an example of how we set up and run an analysis using the RAT toolbox. Every call to the toolbox has two parts: the **problem definition** class, where we define the model, add the data and define our contrasts, and the **controls class** where we tell the toolbox what type of analysis we would like to do. The reason for splitting things up in this way is that once our model is defined, we can interact with it in various ways without needing to modify the model. So, we can experiment with our data, trying out different types of analysis (more of that in chapter 4), and explore the landscape of solutions by simply modifying the *controls* class, leaving the *problem* class alone. 
 
 As well as having two inputs, RAT always provides two outputs, so the call to the toolbox is always of this form:-
```
DrPaulSharp, line 8: Just below here, we discuss using "projectClass", but we now want users to define their problem using the "project" routine instead, so this should be rewritten.

DrPaulSharp (approved):

docs/source/chapter2.rst
StephenNneji: > Just below here, we discuss using "projectClass", but we now want users to define their problem using the "project" routine instead, so this should be rewritten.

Yes there is a lot if "problemDef" class in the docs which should be removed also. Hopefully when the API is stable

PR #147, opened 2023-06-13
Refactors controlsClass

### LINKED ISSUE #145: Refactor the Controls Class
Currently controls are specified as 
 
```
    controls = controlsClass();
    controls.calcSldDuringFit = 'no';
    controls.procedure = 'bayes';
    controls.parallel = 'contrasts';
    controls.maxIter = 1e9;
    controls.repeats = 3;
    controls.nsimu = 5000;
    controls.burnin = 1000;
```

Add a method called `setProcedure` which will receive the procedure name and a cell array with name-value options i.e.

```
controls = controlsClass();
controls.setProcedure(procedures.Dream, {'nSamples', 1e9, 'nChains', 3, 'lambda', 5000, 'parallel', 'contrasts'})
```
The method should accept general options such as 'parallel' and any option that is available for specified procedure type. It should throw an exception for the incorrect option.

Add appropriate unit test and update docs.

RabiyaF:
Closes #145 

**Tasks**
- [x] Remove references to `controlDef()` from docs.
- [x] Update the Matlab LiveScripts to use `controlClass()` instead of `controlDef()`
- [x] Fix the example LiveScripts and add test for the them if possible 
- [x] Add the `setProcedure()` method to controlsClass
- [x] Update all references of controlClass with new method for instantiation.
- [x] Write test for the new method.
- [x] Update docs to explain new usage of method.


docs/source/images/toolbox.png
RabiyaF, line 1: For now this is just a placeholder image. I updated this image to remove the reference to controlsDef() class. The rest of the image is practically the same. However, I understand that it might not be representative of the current architecture. Let me know if you want me to make any changes to it in the future.

tests/testControlsClass.m
```diff
@@ -342,5 +342,240 @@ function testDisplayGroups(testCase)
...
+            testCase.verifyEqual(testCase.controls.resamPars, [0 10], 'setProcedure method is not working');
+            testCase.verifyEqual(testCase.controls.display, displayOptions.Notify.value, 'setProcedure method is not working');
+
+            % Test passing invalid parameter values for dream procedurE
+            testCase.verifyError(@() testCase.controls.setProcedure(procedures.Dream.value,...
+                {'boundHandling', 'invalid'}), invalidOption.errorID);
+            
+            % Test passing wrong parameter for dream procedure
+            testCase.verifyError(@() testCase.controls.setProcedure(procedures.Dream.value,...
+                {'tolX', 1e-6}), invalidParameter.errorID); % Simplex Parameter
```
DrPaulSharp, line 388: I'm not sure of the need for a new custom exception here - I think this fits into the invalidOption exception, which covers possible options that are not valid in a particular case.

DrPaulSharp (commented):
This looks good, I've a few small comments, and I think it's worth considering if you can reduce some of the repeated code in the process input routines - I've made a suggestion on this.

docs/source/api.rst
```diff
@@ -427,7 +435,7 @@ After the user has defined the projectClass and controlsDef class, the user can
 When the RAT function is called, the classes are passed into internal functions like `RatParseClassToStructs_new` which takes the classes and breaks them down into cells, 
 limits,prior and more importantly converts the project class to struct. 
 
-Then, the `RAT_main` function redirects the control flow based on what procedure is selected in controlsDef class. One of the redirecting functions will call the reflectivity_calculation_wrapper
+Then, the `RAT_main` function redirects the control flow based on what procedure is selected in controlsClass. One of the redirecting functions will call the reflectivity_calculation_wrapper
```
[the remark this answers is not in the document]
RabiyaF: I left the other parts of the docs as they are and just made changes to controlsDef/controlsClass. There are also a lot of references to problemDef in docs and images. However, I will make the changes to the typos you spotted here. But I think that a seperate issue might be created to update the docs and fix the other typos.

API/controlsClass/controlsClass.m
```diff
@@ -286,5 +357,313 @@
...
+            try
+                parse(p, properties{:});
+                inputBlock = p.Results;
+            catch ME
+                if (strcmp(ME.identifier,'MATLAB:InputParser:UnmatchedParameter'))
+                    throw(invalidParameter('Only nSamples, nChains, lambda, pUnitGamma, boundHandling, parallel, calcSldDuringFit, resamPars and display can be set while using the DREAM procedure'));
+                else
+                    rethrow(ME)
+                end
+            end
```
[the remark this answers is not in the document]
RabiyaF: I will change the exception part.

I agree there is a lot of repetition (Hence the missed typos in the docs after copy and pasting).  I left the functions like this because parser are defined in a similar way in the other classes. I could not think of a way to cleanly abstract out the addParameter() and the set value step into a more general form that I could then reuse for each of the different procedures. I think it can be done by passing list to functions. But I thought it might become less readable.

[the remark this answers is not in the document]
DrPaulSharp: Yes, I agree with that. Apart from the try-catch block, there was nothing obvious to me that would clearly improve the code whilst also reducing repetition. I also agree it could be done in ways which make the code less readable, which I'd also be inclined to avoid.

API/controlsClass/controlsClass.m
```diff
@@ -664,6 +624,19 @@
             obj.display = inputBlock.display;
         end
 
+        function inputBlock = parseInputs(obj, p, properties, errorMsg)
```
DrPaulSharp, line 627: I'd make this routine private (i.e., include in a `methods(Access = private)` block, since it refers only to the routines defined above and will not be used outside of the class.

RabiyaF: The parsing methods have been made private.

DrPaulSharp (approved):

PR #148, opened 2023-06-16
Updated reflectivity calculation - using Imaginary SLD

arwelHughes:
- New Abeles calculation which correctly reproduces benchmarks
 - Handles Imaginary SLD correctly
 - Introduced a new 'usieImaginary' flag in problemDef
 - Imaginary option handled in Standard Layers and both Custom modes
 - Re-done SLD resampling to correctly handle imaginary SLD
 - Most tests pass, apart from 'callReflectivity' as this now expects the useImaginary flag. This not fixed my makeInputsandOutputs (though the others were)
 - Started modifying examples so that I can include new (imaginary) types, and also validation folder (TODO)


targetFunctions/common/reflectivityCalculations/abeles/abelesParallelPoints.m
StephenNneji, line 1: Can these files be moved to _targetFunctions/common/abeles_ instead of _targetFunctions/common/reflectivityCalculations/abeles_

StephenNneji (approved):

PR #149, opened 2023-06-21
Adds "writeScript" routine

DrPaulSharp:
This routine generates a script to reproduce the project/domains object that is currently defined.

There are a couple of points to consider:

- Can we reduce some duplication (certainly the layers part can be incorporated when NaNs are resolved)?
- Can we neaten up the routine whilst ensuring all cases are accounted for?
- Do we need the two sets of tests?


arwelHughes:
Instead of adding the parameters individually, could we make an array first and then add them together with addParamGroup? This leads to a far more intuitive script for users. 

This script would do that (although to screen rather than file the way it's written here.....)

```
`function params = paramsToScript(problem)

% Converts the parameters table to a printed table for a script

format short
paramsTable = problem.parameters.paramsTable;

[nParams,~] = size(paramsTable);

%Print the params definition:
fprintf('params = { \n');

% Print the table contents row by row:
for i = 1:nParams
    thisName = paramsTable{i,1};
    thisMin = paramsTable{i,2};
    thisVal = paramsTable{i,3};
    thisMax = paramsTable{i,4};
    thisFit = paramsTable{i,5};
    
    if thisFit == 1
        thisFit = 'true';
    else
        thisFit = 'false';
    end
    
    fprintf('{ ''%s'' \t %g \t %g \t %g \t %s \t}; \n',...
        thisName, thisMin, thisVal, thisMax, thisFit)
end

fprintf('};');

end
```

tests/testProjectClass.m
```diff
@@ -553,6 +572,55 @@ function testToStruct(testCase)
...
+            % Populates project properties for the tests
+            testCase.populateProject();
+
+            % Verify error for invalid input and write the script
+            testCase.verifyError(@() testCase.project.writeScript(script="invalid.txt"), invalidValue.errorID)
+            testCase.project.writeScript(objName="problem", script="newScript");
+            run("newScript.m");
+    
+            % Test general properties
+            testCase.verifyEqual(testCase.project.experimentName, problem.experimentName, 'experimentName is not correctly regenerated by the script')
```
StephenNneji, line 588: Inconsistent semicolons

StephenNneji (approved):

tests/testProjectClass.m
DrPaulSharp: Which part do you think exactly? As I see it, the tests do not have semicolons but the lines that execute code do, which is the convention we have used elsewhere.

StephenNneji: I thought we agreed the properties should not have semicolons? Most of the test code seem to have them

PR #153, opened 2023-07-05
Removes definition of layers for custom models

DrPaulSharp:


API/projectClass/projectClass.m
```diff
@@ -188,18 +190,7 @@
             invalidTypeMessage = sprintf('Geometry must be a geometryOptions enum or one of the following strings (%s)', ...
                                          strjoin(geometryOptions.values(), ', '));
             obj.geometry = validateOption(geometry, 'geometryOptions', invalidTypeMessage).value;
-        end
-        
-        function obj = setModelType(obj, modelType)
```
DrPaulSharp, line 193: The "setModelType" routine has been used widely in the past, and is used throughout the examples, so we'll need to discuss its removal carefully. The case for doing so is that the model type sets what is allowed elsewhere, and changing it on the fly could lead to many unintended consequences. For example, a set of contrasts could be defined using layers, with the model type then changed to custom where the layers details are removed and the contrasts have dangling references.

arwelHughes:
But changing on the fly is actually a use-case which would come up quite often. A user might start their analysis project with layers, find they need something more sophisticated and decide to switch to a custom model type (this will be available in GUI functionality for example).

So the desired behaviour at this point is that it would keep the data, backgrounds, contrasts etc that the user has already defined, but the layers *and the layers in the contrasts models* would disappear, and the user would then need to edit in the new model information. All the existing references in the contrasts to backgrounds, resolutions etc would stay.

So in other words, the desired behaviour would be handle the dangling references cleanly, rather than to prevent the switch.

Cheers,
A


From: Paul Sharp ***@***.***>
Date: Wednesday, 5 July 2023 at 16:29
To: RascalSoftware/RAT ***@***.***>
Cc: Subscribed ***@***.***>
Subject: Re: [RascalSoftware/RAT] Removes definition of layers for custom models (PR #153)

@DrPaulSharp commented on this pull request.

In API/projectClass/projectClass.m<https://github.com/RascalSoftware/RAT/pull/153#discussion_r1253284543>:

> @@ -188,18 +190,7 @@

             invalidTypeMessage = sprintf('Geometry must be a geometryOptions enum or one of the following strings (%s)', ...

                                          strjoin(geometryOptions.values(), ', '));

             obj.geometry = validateOption(geometry, 'geometryOptions', invalidTypeMessage).value;

-        end

-

-        function obj = setModelType(obj, modelType)

The "setModelType" routine has been used widely in the past, and is used throughout the examples, so we'll need to discuss its removal carefully. The case for doing so is that the model type sets what is allowed elsewhere, and changing it on the fly could lead to many unintended consequences. For example, a set of contrasts could be defined using layers, with the model type then changed to custom where the layers details are removed and the contrasts have dangling references.

StephenNneji (approved):

PR #158, opened 2023-09-21
Fixes trigger event and interface so it works properly with domains

StephenNneji:
- The modifies the event system so it works with domains.
- The eventDemo example is updated
- The plotSLD function is unified to handle standard and domains


API/events/triggerEvent.m
```diff
@@ -36,22 +36,35 @@ function triggerEvent(eventType, data)
...
+   
+                switch problemDef.TF
+                    case 'domains'
+                        [sldProfiles, nSldProfiles] = packCellArray(result{5}, 1);
+                        [sldProfiles2, nSldProfiles2] = packCellArray(result{5}, 2);
+                        [layers, nLayers] = packCellArray(result{6}, 1);
+                        [layers2, nLayers2] = packCellArray(result{6}, 2);
+                    otherwise
+                        [sldProfiles, nSldProfiles] = packCellArray(result{5}, 1);
+                        [layers, nLayers] = packCellArray(result{6}, 1);
```
DrPaulSharp, line 52: These two lines are also included in the previous clause. Presumably they can be moved out of the switch case?

examples/eventDemo.m
```diff
@@ -16,3 +17,69 @@
...
+problem.setDomainContrastModel(1,'Domain1 Layer');
+problem.setDomainContrastModel(2,'Domain2 Layer');
+
+problem.addContrast('name','Domain Test',...
+    'background','Background 1',...
+    'resolution','Resolution 1',...
+    'scalefactor', 'Scalefactor 1',...
+    'resample',false,...
+    'nba', 'SLD Air',...        % This is bulk out
+    'nbs', 'SLD D2O',...        % This is bulk in
```
DrPaulSharp, line 71: Howling at the moon here, but why on earth were these variable names chosen?

DrPaulSharp: Isn't nba bulk_in? It seems that way in `projectClass.m`

StephenNneji: Nice catch, I copied the example from standardTest.m, its most likely a bad comment. I will fix in both places

API/events/triggerEvent.m
StephenNneji: yes, will do

PR #160, opened 2023-10-04
Adds unit tests for domains reflectivity calculation

StephenNneji:
- Adds unit tests for domains reflectivity calculation
- Fixes incorrectly names target functions
- Fixes bugs where customXY function for domains were not returning outsubs

Note
- outsubs and allroughs are meant to be the same in the domains target function according to @arwelHughes this will be addressed later


tests/domainsTFReflectivityCalculation/domainsCustomXY.m
```diff
@@ -0,0 +1,61 @@
...
+    
+    % Set the scalefactor...
+    problem.setScalefactor(1,'Value',1,'min',0.5,'max',2,'fit',true);
+    
+    
+    % Set the bulk SLD
+    problem.setBulkIn(1,'name','Air','Value',0,'fit',false);
+    
+    % and scalefactor...
+    problem.setScalefactor(1,'min',0.8,'Value',1,'max',1.1);
```
DrPaulSharp, line 42: The same scalefactor is modified twice in the space of eight lines. Is this right?

DrPaulSharp (approved):

tests/domainsTFReflectivityCalculation/domainsCustomXY.m
StephenNneji: Nice catch, I lifted this from an existing example so I will fix here and in the example

PR #173, opened 2023-11-22
Replaces the "callOutputAndPlotFcns" in the simplex with the new plot events

### LINKED ISSUE #168: Compile throws a warning in drawMultiNest.m
Warning: Index variable k is used outside its corresponding FOR loop. The run-time behavior will differ from MATLAB if the loop is
not executed.

The offending code  is 
                
```
        for k=1:K
            if rval < fracvol(k)
                continue
            else
                break
            end
        end
        k0 = k;  <<<<<<<<<<<< k is used outside the loop
       
        % extract bounding matrix and centroid for that ellipsoid
        B = Bs((k0-1)*ndims+1:k0*ndims,:);
```
possible fix

```
        k0 = 0  <<<<<<<<<<<< is this a good initial value?
        for k=1:K
            k0 = k
            if rval < fracvol(k)
                continue
            else
                break
            end
        end
       
        % extract bounding matrix and centroid for that ellipsoid
        B = Bs((k0-1)*ndims+1:k0*ndims,:);
```

StephenNneji:
`callOutputAndPlotFcns` is used 5 times in **fMinSearch.m** these have now been replaced with `triggerEvents`. Also adds a method `hasPlotHandler` so `triggerEvent` can check when a plothandler is present and exit early if not.

Also Fixes #168


StephenNneji:
@arwelHughes I am concerned this could affect the speed significantly when plotting during optimization 

```
problem = load('DSPCBilayerProjectClass.mat').thisProjectClass;

% Make a controls block
controls = controlsClass();
controls.parallel = parallelOptions.Points;
controls.procedure = procedures.Simplex;
controls.maxIter = 1;

figure(2); clf
eventManager.register(eventTypes.Plot, 'plotRefSLDHelper')
% eventManager.register(eventTypes.Message, 'disp')

[problem1, result1] = RAT(problem, controls);

eventManager.clear()
```
When running the example above, the times go from ~0.02 sec without plotting to ~1 sec with plotting. This time difference will grow with more iterations. Potential solutions

- Reduce number of calls to plot from 5 to maybe 1
- Speedup of `plotRefSLDHelper` since plot cannot be performed on another thread

DrPaulSharp (approved):

PR #177, opened 2023-11-26
Added tests for mergeStructs

RabiyaF:
Added tests for the mergeStructs function.

### DIFF
--- a/tests/testMergeStructs.m
+++ b/tests/testMergeStructs.m
@@ -1,0 +1,21 @@
+classdef testMergeStructs < matlab.unittest.TestCase   
+
+    properties (TestParameter)
+        s = {struct('s1', 1, 's2', 2),...
+             struct('s1', 1),...
+             struct('s1', 1, 's2', 2, 's3', 3)};
+        d = {struct('d1', 1, 'd2', 2),...
+             struct('d1', 1),...
+             struct('d1', 1, 'd2', 2)};
+        r = {struct('s1', 1, 's2', 2, 'd1', 1, 'd2', 2),...
+             struct('s1', 1, 'd1', 1),...
+             struct('s1', 1, 's2', 2, 's3', 3, 'd1', 1, 'd2', 2)};
+    end
+
+    methods (Test, gitParameterCombination="sequential")
+        function testMergeResults(testCase, s, d, r)
+            result = mergeStructs(s, d);
+            testCase.verifyEqual(result, r);
+        end
+    end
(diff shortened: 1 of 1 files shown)


tests/testMergeStructs.m
```diff
@@ -0,0 +1,21 @@
...
+             struct('s1', 1, 's2', 2, 's3', 3)};
+        d = {struct('d1', 1, 'd2', 2),...
+             struct('d1', 1),...
+             struct('d1', 1, 'd2', 2)};
+        r = {struct('s1', 1, 's2', 2, 'd1', 1, 'd2', 2),...
+             struct('s1', 1, 'd1', 1),...
+             struct('s1', 1, 's2', 2, 's3', 3, 'd1', 1, 'd2', 2)};
+    end
+
+    methods (Test, gitParameterCombination="sequential")
```
DrPaulSharp, line 15: Remove the word "git" here.

tests/testMergeStructs.m
```diff
@@ -0,0 +1,21 @@
...
+        s = {struct('s1', 1, 's2', 2),...
+             struct('s1', 1),...
+             struct('s1', 1, 's2', 2, 's3', 3)};
+        d = {struct('d1', 1, 'd2', 2),...
+             struct('d1', 1),...
+             struct('d1', 1, 'd2', 2)};
+        r = {struct('s1', 1, 's2', 2, 'd1', 1, 'd2', 2),...
+             struct('s1', 1, 'd1', 1),...
+             struct('s1', 1, 's2', 2, 's3', 3, 'd1', 1, 'd2', 2)};
+    end
```
DrPaulSharp, line 13: I think it's worth adding a test case where the structs have different values for the same field.

DrPaulSharp (approved):

tests/testMergeStructs.m
RabiyaF: Ohh It might have happened when I was committing it 😄

tests/testMergeStructs.m
RabiyaF: It will raise an error. I did try that yesterday but then I removed it. The approach used to combine structs is the first one [here](https://www.mathworks.com/matlabcentral/answers/96973-how-can-i-concatenate-or-merge-two-structures). The underlying assumption of the approach is that there will be no duplication of field names. But if that is a possibility then we need to define a policy of how to combine. So, that is why I left it simple.

PR #179, opened 2023-11-29
Refactors contrasts class and struct

DrPaulSharp:


StephenNneji (approved):
There are still a few uses of `nba`, `nbs` in the targetFunctions. Also `cNbas`, `cNbss`

DrPaulSharp:
> There are still a few uses of `nba`, `nbs` in the targetFunctions. Also `cNbas`, `cNbss`

Yeah. This follows the extraction of individual parameters from the structs in `extractProblemParams`. I thought best to go one step at a time.

Issue #181, opened 2023-11-30
Renaming variable causes tests to fail


DrPaulSharp:
When renaming "nbs" as "bulkOut" in "targetFunctions/+nonPolarisedTF/*/processCustomFunctions.m", the tests fail with the error: 

"Index exceeds the number of array elements. Index must not exceed 1." in `backSort.m`.

Is there a reason why the input and output to `backSort.m` must be named differently - I not the output has been named as "bulkOut" historically, unlike the rest of the `targetFunctions` code.

arwelHughes:
Which line number? Have you tried running the Matlab version with breakpoints?

StephenNneji:
The problem is that when the nba is renamed to bulkOut, the value now changes in the for loop leading to the index error in subsequent iterations
```
for i = 1:numberOfContrasts     % TODO - the ambition is for parfor here, but would fail for Matlab and Python CM's..
         ...
         [~,~,~,bulkIn,bulkOut,~] = backSort(..., bulkIn,bulkOut,res);
         ...
    end
```
 The solution is to change variable names of the outputs from backsort

DrPaulSharp:
Ahh, of course. I think a bit of code blindness set in there. I'll tidy that up shortly then.

Cheers,
Paul


Dr Paul Sharp

Senior Software Scientist

ISIS Neutron and Muon Source

STFC Rutherford Appleton Laboratory

Harwell, Oxford

OX11 0QX

[cid:7699df67-7354-4c5e-81b4-83be4ad76997]

________________________________

arwelHughes:
Also, bulkIn going into backsort is an array, whereas in backsort a single value is chose (cNba's etc are arrays of indexes). So, the array / single value definition will have been made upstream somewhere....

DrPaulSharp:
Yeah, that's where I thought the problem might have been, and hence what I was worried about.

Issue #175, opened 2023-11-24
Redundant/commented out code


DrPaulSharp:
The use of the following routines in the code is not clear (calls to them are either commented out or not present). I need a hand working out what they are used for. If they are not required, they should be deleted.

`compile/reflectivityCalculation/reflectivityCalculationCodeOnlyScript.m` -- is required
~~`compile/reflectivityCalculation/setCompilePaths.m`~~ - Removed in #255 

~~`minimisers/DE/runDE.m` - function "PlotIt"~~ - Removed in #178 

~~`minimisers/DREAM/lowerCaseStruct.m`~~ - Removed in #357 
`minimisers/DREAM/mvnorpf.m`
`minimisers/DREAM/Diagnostics/chisRnd.m`
`minimisers/DREAM/functions/DREAM.m` - has this been replaced by `ratDREAM.m`?
~~`minimisers/DREAM/functions/evaluateRATModel.m`~~ - Removed in #178 
`minimisers/DREAM/functions/GL.m`
~~`minimisers/DREAM/functions/initializeDREAMForRAT.m`~~ - Removed in #178 
`minimisers/DREAM/functions/measError.m`
~~`minimisers/DREAM/functions/setupDREAMForRAT.m`~~ - Removed in #178 
`minimisers/DREAM/functions/whittle.m`
`minimisers/DREAM/postprocessing/DREAMPostprocessing.m`
`minimisers/DREAM/postprocessing/genParSet.m`
`minimisers/DREAM/postprocessing/predMCMC.m`

~~`minimisers/generalUtils/checkBounds.m`~~ - Removed in #357 
~~`minimisers/generalUtils/bayesStats/calcMCMCstatRefErrors.m`~~ - Removed in #357 
~~`minimisers/generalUtils/bayesStats/confIntervals.m`~~ - Removed in #357 
~~`minimisers/generalUtils/bayesStats/packPriors.m`~~ - Removed in #356 
~~`minimisers/generalUtils/bayesStats/ratInputs.mat`~~ - Removed in #357 
~~`minimisers/generalUtils/bayesStats/ratTest.m`~~ - Removed in #357 

~~`minimisers/NSMain/loopCell.m`~~ - Removed in #357 
~~`minimisers/NSMain/parseBayesResults.m`~~ - Removed in #357 
~~`minimisers/NSMain/plotLivePointsWithEllipses.m`~~ -- Removed in #264 

[...]

PR #182, opened 2023-11-30
Unify variable naming convention

DrPaulSharp:
The use of "nba" and "nbs" is retained in "processCustomFunction" routines due to the bug identified in #181


minimisers/generalUtils/checkBounds.m
```diff
@@ -24,7 +24,7 @@
 
 fitWhichNbas = problem.fitBulkIn;
 fittingNbas = problem.nba(find(fitWhichNbas));
-fittingNbaConstr = problem.nbairs_constr(find(fitWhichNbas),:);
+fittingNbaConstr = problem.bulkIns_constr(find(fitWhichNbas),:);
```
StephenNneji, line 27: please remove the underscore

targetFunctions/common/shiftData.m
```diff
@@ -5,7 +5,7 @@
 % INPUTS:
 %
 %     * scalefac = problem.scalefactors;
```
StephenNneji, line 7: Please rename `shifted_data` and `scalefac`

StephenNneji (approved):

minimisers/generalUtils/checkBounds.m
DrPaulSharp: I'll get on it, but we should be removing this file, as per #175.

PR #195, opened 2024-01-19
Modifies the CPP custom layer interface to use std::vector

StephenNneji:
- Fixes a bug in processCustomFunction
- Modifies C++ custom function interface to use std::vector instead of pointers

### DIFF
--- a/targetFunctions/+domainsTF/+customLayers/processCustomFunction.m
+++ b/targetFunctions/+domainsTF/+customLayers/processCustomFunction.m
@@ -39,8 +39,8 @@ function [allLayers,allRoughs] = processCustomFunction(contrastBackgrounds,contr
             [thisContrastLayers1, allRoughs(i)] = callMatlabFunction(params,i,functionHandle,thisBulkIn,thisBulkOut,numberOfContrasts,1);
             [thisContrastLayers2, ~] = callMatlabFunction(params,i,functionHandle,thisBulkIn,thisBulkOut,numberOfContrasts,2);
         else
-            [thisContrastLayers1, allRoughs(i)] = callCppFunction(params,thisBulkIn,thisBulkOut,i,0,functionHandle);
-            [thisContrastLayers2, ~] = callCppFunction(params,thisBulkIn,thisBulkOut,i,1,functionHandle);
+            [thisContrastLayers1, allRoughs(i)] = callCppFunction(params,thisBulkIn,thisBulkOut,i-1,0,functionHandle);
+            [thisContrastLayers2, ~] = callCppFunction(params,thisBulkIn,thisBulkOut,i-1,1,functionHandle);
         end
 
         % If the output layers has 5 columns, then we need to do
--- a/targetFunctions/+domainsTF/+customLayers/processCustomFunction.m
+++ b/targetFunctions/+domainsTF/+customLayers/processCustomFunction.m
@@ -36,11 +36,11 @@ function [allLayers,allRoughs] = processCustomFunction(contrastBackgrounds,contr
         coder.varsize('thisContrastLayers2',[10000, 6],[1 1]);
 
         if isnan(str2double(functionHandle))
-            [thisContrastLayers1, allRoughs(i)] = callMatlabFunction(functionHandle, params, thisBulkIn, bulkOut, i, 1);
-            [thisContrastLayers2, ~] = callMatlabFunction(functionHandle, params, thisBulkIn, bulkOut, i, 2);
+            [thisContrastLayers1, allRoughs(i)] = callMatlabFunction(functionHandle, params, thisBulkIn, bulkOut(contrastBulkOuts), i, 1);
+            [thisContrastLayers2, ~] = callMatlabFunction(functionHandle, params, thisBulkIn, bulkOut(contrastBulkOuts), i, 2);
         else
-            [thisContrastLayers1, allRoughs(i)] = callCppFunction(functionHandle, params, thisBulkIn, bulkOut, i-1, 0);
-            [thisContrastLayers2, ~] = callCppFunction(functionHandle, params, thisBulkIn, bulkOut, i-1, 1);
+            [thisContrastLayers1, allRoughs(i)] = callCppFunction(functionHandle, params, thisBulkIn, bulkOut(contrastBulkOuts), i-1, 0);
+            [thisContrastLayers2, ~] = callCppFunction(functionHandle, params, thisBulkIn, bulkOut(contrastBulkOuts), i-1, 1);
         end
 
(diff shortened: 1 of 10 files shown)


examples/customLayers/customBilayer.cpp
```diff
@@ -70,40 +70,40 @@ extern "C" {
 
         // Manually deal with hydration for layers in
         // this example.
-        double oxSLD = (oxideHydration * *bulkOut) + ((1 - oxideHydration) * oxideSLD);
-        double headSLD = (headHydration * *bulkOut) + ((1 - headHydration) * SLDhead);
-        double tailSLD = (bilayerHydration * *bulkOut) + ((1 - bilayerHydration) * SLDtail);
+        double oxSLD = (oxideHydration * bulkOut[0]) + ((1 - oxideHydration) * oxideSLD);
```
StephenNneji, line 73: @Arwel the python version of this example does indexes the bulkOut with contrast number i.e `bulkOut[contrast]` but this is not the case in the C++ version. Is this intentional?

arwelHughes:
I was just looking at that - no it isn't!

Bulk out should come in as an array of all the bulk_outs for the given contrasts - mainly for backwards comptaibility but also there are (rare) occasions where the SLD of the previous contrast may be useful to know. So, if bulk_out is coming in as an array it shuld be bulk_out[contrast] (or whatever) in C also...

btw - I very much like this! ...
       // Make the layers
        // oxide...
        output.push_back(oxideThick);
        output.push_back(oxSLD);
        output.push_back(subRough);
🙂

A
p.s.
Can we remove the debug 'std::cout... ' also?


examples/customLayers/customBilayer.cpp
Arwel: Wrong Arwel I'm afraid, you're after @arwelHughes. Been far too long since I did any c++

StephenNneji: Apologies @Arwel my bad

arwelHughes:
> @arwelHughes On closer inspection, the bulkOut is not passed as an array in the C++ apparently `backsort` does the indexing in `processCustomFunction`
> 
> The reason the MATLAB version receives an array is because of these lines in `callMatlabFunction`
> 
> ```
> if coder.target('MATLAB')
>     % For backwards compatability with Rascal1...
>     bulkOuts = zeros(nContrasts,1);
>     bulkOuts(contrast) = bulkOut;
> ```
> 
> Should we be passing `contrastBulkIns` and `contrastBulkOuts` into these function instead?

Well, as the comment says, if we just pass the 'per contrast' value of bulkOut we will essentially break 'r1ToProjectClass' for custom models, which for user engagement might be a bad idea initially. It's logical to do so though.  I think we should persist with backwards compatibility initially and for a while, but the pre-announce the change for a future release.

But, there is actually no reason to do that for C custom models (they aren't available for RasCAL1, so no BC issues...). However, making C custom models using Coder source code generation (which I'm going to encourage) would be affected, sooo ... we should discuss at the next meeting.
A

examples/customLayers/customBilayer.cpp
```diff
@@ -70,49 +65,44 @@ extern "C" {
 
         // Manually deal with hydration for layers in
         // this example.
-        double oxSLD = (oxideHydration * *bulkOut) + ((1 - oxideHydration) * oxideSLD);
-        double headSLD = (headHydration * *bulkOut) + ((1 - headHydration) * SLDhead);
-        double tailSLD = (bilayerHydration * *bulkOut) + ((1 - bilayerHydration) * SLDtail);
+        double oxSLD = (oxideHydration * bulkOut[contrast]) + ((1 - oxideHydration) * oxideSLD);
+        double headSLD = (headHydration * bulkOut[contrast]) + ((1 - headHydration) * SLDhead);
+        double tailSLD = (bilayerHydration * bulkOut[contrast]) + ((1 - bilayerHydration) * SLDtail);
```
DrPaulSharp, line 70: How does this code work? At a glance multiplying by a pointer and a single element of a vector seem to be completely different operations.

targetFunctions/+domainsTF/+customLayers/processCustomFunction.m
```diff
@@ -36,11 +36,11 @@
         coder.varsize('thisContrastLayers2',[10000, 6],[1 1]);
 
         if isnan(str2double(functionHandle))
-            [thisContrastLayers1, allRoughs(i)] = callMatlabFunction(params,i,functionHandle,thisBulkIn,thisBulkOut,numberOfContrasts,1);
-            [thisContrastLayers2, ~] = callMatlabFunction(params,i,functionHandle,thisBulkIn,thisBulkOut,numberOfContrasts,2);
+            [thisContrastLayers1, allRoughs(i)] = callMatlabFunction(functionHandle, params, thisBulkIn, bulkOut, i, 1);
+            [thisContrastLayers2, ~] = callMatlabFunction(functionHandle, params, thisBulkIn, bulkOut, i, 2);
```
DrPaulSharp, line 40: Why change from "thisBulkOut" to "bulkOut" but not "thisBulkIn" to "bulkIn"?

DrPaulSharp: Since both are changed to vectors above.

DrPaulSharp (approved):
Looks fine, just a couple of questions from me.

targetFunctions/+domainsTF/+customLayers/processCustomFunction.m
StephenNneji: `thisBulkOut` contains only a single value i.e `bulkOut[contrast]` but according to Arwel the custom function should receive the full array to maintain backward compatibility with RasCAL

examples/customLayers/customBilayer.cpp
StephenNneji: In the original code `(bilayerHydration * *bulkOut)` is equivalent to `(bilayerHydration * bulkOut[0])` the former used pointer dereferencing to get the first value in the array. This worked because we previously passed the bulk out for the current contrast only so the first value was always `bulkOut[contrast]` but now we pass the full array

targetFunctions/+domainsTF/+customLayers/processCustomFunction.m
DrPaulSharp: Makes sense, so why not do the same with bulk in?

StephenNneji: My guess is that the bulkIn was not passed like that in RasCAL. Is this correct @arwelHughes ?

PR #201, opened 2024-01-29
Introduces ability to change number of threads in tests

DrPaulSharp:
Default behaviour is the same as before, but "numThreads" can easily be set to a specific number.


StephenNneji:
Closing for now, can be reopened later if needed

PR #202, opened 2024-02-06
Refactors event manager to use enum class and std::function

StephenNneji:
Refactors event manager to use c++ enum class and std::function callback instead of the c-style version used before. This would make it easier for the pybind extension implementation

### DIFF
--- a/compile/events/eventManagerInterface.cpp
+++ b/compile/events/eventManagerInterface.cpp
@@ -1,6 +1,7 @@
 #include "mex.h"
 #include <cstring>
 #include <memory>
+#include <functional>
 #include "eventManager.h"
 #include "dylib.hpp"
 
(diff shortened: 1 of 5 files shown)


compile/events/eventManagerInterface.cpp
```diff
@@ -136,16 +137,14 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[])
...
-            if (type == eventType){
-   	            void (*cbPtr)(const baseEvent&) = eventCallback;
-                initDyLib();
-                auto addListener = library->get_function<void(enum eventTypes, void (*)(const baseEvent&))>("addListener");
-                addListener((enum eventTypes)eventType, cbPtr);
-                return;
-            }
-        }
-        mexErrMsgTxt("Event type not recognized.");
+       if (eventType < 0 || eventType > static_cast<int>(EventTypes::Plot)) {
```
DrPaulSharp, line 140: Is there not a better test here? It's very easy to add to/modify this enum down the line and plot may not be the final entry. Is there a way of obtaining the number of options?

DrPaulSharp (approved):

compile/events/eventManagerInterface.cpp
StephenNneji: To the best of my knowledge, there is no way to get the count. One solution I've seen is to add an extra enum e.g. END which marks the end and will not be moved

DrPaulSharp: Hmmm, not spectacular either way then.

StephenNneji: I will revisit if I find a better solution

PR #203, opened 2024-02-12
Adds action to deploy generated C++ code to a separate branch

StephenNneji:
- The _cppDeploy_ script uses the  _packNGo_ function to get all files required to build the cpp and adds the event files also
- Theses files are deployed to the **generated_source** branch by the Github action
- The OS specific nightly releases i.e. Linux.zip, Windows.zip now include the whole source not just the **compile** folder

### DIFF
--- a/.github/workflows/runTests.yml
+++ b/.github/workflows/runTests.yml
@@ -55,11 +55,11 @@ jobs:
       with:
         ref: generated_source
         path: rat_source
-    - name: Deploy Souce
+    - name: Deploy Source
       run: |
         wget https://github.com/RascalSoftware/RAT/releases/download/nightly/Linux.zip
         unzip Linux.zip
-        rm -rf rat_source/*.c rat_source/*.cpp rat_source/*.h
+        rm -rf rat_source/*.c rat_source/*.cpp rat_source/*.h rat_source/*.hpp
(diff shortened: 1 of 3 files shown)


.github/workflows/runTests.yml
```diff
@@ -42,5 +42,28 @@ jobs:
...
+    - name: Checkout Source
+      uses: actions/checkout@v4
+      with:
+        ref: generated_source
+        path: rat_source
+    - name: Deploy Souce
+      run: |
+        wget https://github.com/RascalSoftware/RAT/releases/download/nightly/Linux.zip
+        unzip Linux.zip
+        rm -rf rat_source/*.c rat_source/*.cpp rat_source/*.h
```
DrPaulSharp, line 62: Have you accounted for the files: "classHandle.hpp"  "dylib.hpp"  "eventHelper.hpp" & "rtw_proj.tmw". They are included in cppDeploy but not deleted here. Is that ok?

DrPaulSharp (approved):

.github/workflows/runTests.yml
StephenNneji: I will update this line for .hpp but remove rtw_proj.tmw from the Matlab

PR #213, opened 2024-03-08
Uses enums in the coder generated code

StephenNneji:


3rdParty/mcmcstat/runBayes.m
```diff
@@ -67,14 +67,14 @@
 results = [];
 loop = int32(loop);
 for i = 1:loop
-    if ~ strcmpi(display,'off')
+    if ~ strcmpi(display, coderEnums.displayOptions.Off)
```
DrPaulSharp, line 70: Should there be a space between "~" and strcmpi here? Will it cause a problem?

DrPaulSharp: It seems ok, but probably best to remove the space anyway.

3rdParty/mcmcstat/runBayes.m
```diff
@@ -67,14 +67,14 @@
...
         fprintf('Running loop %d of %d ',i,loop);
     end
 
     %[results,chain,s2chain,sschain] = mcmcrun_compile(model, data, problem, params, options, results, display);
     [results,chain,s2chain,sschain] = mcmcrun(model, data, params, options, results);
     
-    if ~ strcmpi(display,'off')
+    if ~ strcmpi(display, coderEnums.displayOptions.Off)
```
DrPaulSharp, line 77: and here

API/enums/calculationTypes.m
```diff
@@ -1,11 +1,15 @@
...
         function v = values()
-            v = customEnum.values('calculationTypes');
+            v = customEnum.values(mfilename('class'));
         end
         
         function e = fromValue(value)
-            e = customEnum.fromValue('calculationTypes', value);
+            e = customEnum.fromValue(mfilename('class'), value);
         end
     end
```
DrPaulSharp, line 14: Can we define a (say) "RATEnum" with these (now identical) routines and inherit for each specific enum?

targetFunctions/+domainsTF/+customLayers/single.m
```diff
@@ -121,7 +121,7 @@
     % Now call the core layers reflectivity calculation
     % In this case we are single cored, so we do not parallelise over
     % points
-    parallelPoints = 'single';
+    parallelPoints = coderEnums.parallelOptions.Single;
```
DrPaulSharp, line 124: Remove this line now we don't need the `parallelPoints` variable

DrPaulSharp (commented):
All looks good, just a bit of tidying and a question of inheritance

API/enums/calculationTypes.m
StephenNneji: I am not sure I fully get your suggestion but sadly mfilename will only work in the current file it will not work with inheritance or any other trick I have tried so far

DrPaulSharp: Righto, I thought that might be the problem. If you look at each of the enums the three methods are identical so it seems natural to ask about inheritance.

StephenNneji: Yeah, I will revisit if I find a better solution

DrPaulSharp (approved):

PR #215, opened 2024-03-11
Updated the documentation

arwelHughes:
- Updated docs
 - Added 'ratDocs' function
 - Some debugging of plotting routines


utilities/plotting/plotBayes.m
```diff
@@ -3,13 +3,13 @@ function plotBayes(problem, results)
     figure(10); clf; plotRefSLD(problem,results)
     
     figure(30); clf;
-    bayesShadedPlot(problem,results,'fit','average','KeepAxes',true,'interval',65,'q4',false)
+    bayesShadedPlot(problem,results,'fit','average','KeepAxes',true,'interval',95,'q4',false)
```
StephenNneji, line 6: This value went from 95 to 65 and now back to 95, it might be better to provide these as options to the function

docs/source/images/userManual/calcTypes/domainContrastTable.png
StephenNneji, line 1: We need to think of a way to generated these images automatically so they are always updated

docs/source/DE.rst
```diff
@@ -0,0 +1,67 @@
...
+DE optimizes a problem by maintaining a population of candidate solutions and creating new candidate solutions by combining existing ones according to its simple formulae, 
+and then keeping whichever candidate solution has the best score or fitness on the optimization problem at hand. In this way, the optimization problem is treated as a black box 
+that merely provides a measure of quality given a candidate solution and the gradient is therefore not needed.
+
+As with all the RAT algorithms, DE is selected using the 'procedure' attribute of the controls block:-
+
+.. code:: MATLAB
+
+    controls = controlsClass();
+    controls.procedure = 'DE'
```
StephenNneji, line 21: I would prefer in these tutorials we encourage the users to use the enums

StephenNneji (approved):

PR #217, opened 2024-03-14
Combines serial and parallel routines

### LINKED ISSUE #216: Potential bug in domains standard layers
In `targetFunctions\+domainsTF\+standardLayers` for each of the single and parallel versions we define (lines 19-20):

```
% Additionally extract the additional domain layers details
domainContrastLayers = problemCells{19};
```

Then within the loop over contrasts, we have (lines 116-117):

```
        thisContrastLayers1 = allocateLayersForContrast(domainContrastLayers{1},outParameterisedLayers,useImaginary);
        thisContrastLayers2 = allocateLayersForContrast(domainContrastLayers{2},outParameterisedLayers,useImaginary);
```

Shouldn't this be `domainContrastLayers{i,1}` and `domainContrastLayers{i,2}` where `i` is the loop index? This makes more sense given the structure of the input array and the non-polarised version.

Let me know @arwelHughes.

DrPaulSharp:
Closes #216 and #218


targetFunctions/+domainsTF/customLayers.m
```diff
@@ -0,0 +1,201 @@
...
+        calcAllLayers1{i} = calcAllLayers{i,1};
+        calcAllLayers2{i} = calcAllLayers{i,2};
+    end
+    
+    if strcmpi(parallel, coderEnums.parallelOptions.Contrasts)
+    
+        % Parallel over all contrasts
+        parfor i = 1:numberOfContrasts
+        
+            [backgroundParams(i),qzshifts(i),scalefactors(i),bulkIns(i),...
```
StephenNneji, line 101: Is there a better way to align these functions for readability?

StephenNneji (approved):

targetFunctions/+domainsTF/customLayers.m
DrPaulSharp: Good question. I'm not sure. You could put a space between each variable, but they're already very long.

DrPaulSharp: Hmm, I think if we put the second square bracket on a new line that helps. Give me a mo.

DrPaulSharp: How does that look? I think it's a bit better.

StephenNneji: Yes it better, a bit easier to find the function name

PR #223, opened 2024-04-26
Removes code for qzshifts and undefined layers

DrPaulSharp:


API/projectClass/projectClass.m
```diff
@@ -41,10 +40,14 @@
         protectedParameters
     end
 
+    properties (SetAccess = immutable, Hidden)
+        qzshifts = parametersClass('Qz shift 1',-1e-4,0,1e-4,false,priorTypes.Uniform,0,Inf)
+    end
+
     properties(Access = protected, Constant, Hidden)
         classes = struct(name = ["parameters", "bulkIn", "bulkOut", "scalefactors", "qzshifts", "backgroundParams", "resolutionParams", "domainRatio", "layers", "customFile", "backgrounds", "resolutions", "data", "contrast"], ...
                          addRoutine = ["addParameter", "addBulkIn", "addBulkOut", "addScalefactor", "addQzshift", "addBackgroundParam", "addResolutionParam", "addDomainRatio", "addLayer", "addCustomFile", "addBackground", "addResolution", "addData", "addContrast"], ...
```
StephenNneji, line 49: You removed addQzshift and removeQzshift but left them here. Is this intended?

StephenNneji (approved):

API/projectClass/projectClass.m
DrPaulSharp: It doesn't harm to leave them there since they're only there for reference but probably is clearer to remove them.

PR #225, opened 2024-05-02
Adds "functionName" field to "customFileClass"

DrPaulSharp:


API/projectClass/customFileClass.m
```diff
@@ -74,26 +75,39 @@ function delete(obj)
...
                         newFile = char(inputs{2});
+                        [~, functionName, ~] = fileparts(newFile);
 
-                        newRow = {newName, newFile, supportedLanguages.Matlab.value, ''};
+                        newRow = {newName, newFile, functionName, supportedLanguages.Matlab.value, ''};
                         
                     case 4
 
-                        % Four inputs = assume all inputs supplied
+                        % Four inputs = assume all inputs except function name supplied
```
StephenNneji, line 92: I think this design where argument 3 changes from **function name** to **language** will be confusing to some users. Its always best to add optional args at the end so we could either remove the number 4 case or move function name to the end
```
customFiles.addCustomFile('New Row', 'file.py', 'python', 'C:/stuff')
customFiles.addCustomFile('New Row', 'file.py', 'py_function', 'python', 'C:/stuff')
```

DrPaulSharp: Yeah, that sounds good. I did want to have the function name next to the file name though, but it's probably best to just put it on the end and have everything the same in that case.

DrPaulSharp: Ok, let's do that and see how it goes.

StephenNneji (approved):

PR #228, opened 2024-05-09
Adds set resolution routines

### LINKED ISSUE #226: 'setResolution' seems to be missing from projectClass...
As described in the title....

There is 'setResolutionParam', but there is only add/remove for resolution itself. 
The only option seems to be to use the 'removeResolution' method, followed by 'addResolution' to make a new one.

The methods already exist in resolutionsClass, but are missing upstream (so yes, you can 'problem.resolution.setResolution', but this isn't best?)

A

DrPaulSharp:
Closes #226

### DIFF
--- a/API/projectClass/resolutionsClass.m
+++ b/API/projectClass/resolutionsClass.m
@@ -173,6 +173,17 @@ classdef resolutionsClass < handle
                 obj.resolutions.setValue(row, i + 2, value);
             end
         end
+
+        function obj = setResolutionName(obj, row, name)
+            % Sets the name of a given resolution in the table. Expects 
+            % an index and the new name. 
+            %
+            % resolution.setResolutionName(1, 'new_name');
+            if ~isText(name)
+                throw(exceptions.invalidType(sprintf('%s must be a character array or string', name)));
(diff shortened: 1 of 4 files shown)


API/RAT.m
```diff
@@ -2,6 +2,11 @@
 
 [problemStruct,problemCells,problemLimits,priors,controls] = parseClassToStructs(project,controls);
 
+%for i=1:20
```
StephenNneji: Please remove comments

API/projectClass/resolutionsClass.m
```diff
@@ -173,6 +173,17 @@ function setResolution(obj, row, varargin)
...
             end
         end
+
+        function obj = setResolutionName(obj, row, name)
+            % Sets the name of a given resolution in the table. Expects 
+            % an index and the new name. 
+            %
+            % resolution.setResolutionName(1, 'new_name');
+            if ~isText(name)
+                throw(exceptions.invalidType(sprintf('%s must be a character array or string', name)));
```
StephenNneji, line 183: This line is not called by the tests

PR #236, opened 2024-06-13
Adds code to ensure parameter array indices are valid

DrPaulSharp:


targetFunctions/+domainsTF/customLayers.m
```diff
@@ -17,11 +17,24 @@
     % Extract individual parameters from problemStruct
     [numberOfContrasts, geometry, contrastBackgroundIndices, contrastQzshiftIndices,...
      contrastScalefactorIndices, contrastBulkInIndices, contrastBulkOutIndices,...
-     contrastResolutionIndices, contrastDomainRatioIndices, backgroundParamArray,...
+     contrastResolutionParamIndices, contrastDomainRatioIndices, backgroundParamArray,...
      qzshiftArray, scalefactorArray, bulkInArray, bulkOutArray, resolutionParamArray,...
      domainRatioArray, dataPresent, nParams, params, ~, resample,...
      contrastBackgroundActions, cCustFiles, useImaginary] = extractProblemParams(problemStruct);
     
+    checkIndices(contrastBackgroundIndices, contrastQzshiftIndices,...
```
StephenNneji, line 25: Is it possible to do this check earlier maybe in `parseClassToStructs`? This should reduce duplication and any chance of affecting performance. The function can take problemStruct as input also

DrPaulSharp: I was cautious owing to the fact that the reflectivity calculation is performed numerous times within the minimisers, with changes to problemStruct. Having looked through carefully though, it seems like the changes made in unpackParams cannot affect the length of the parameter arrays, so we should be ok. Let me know if you can think of anything that might be affected though.

API/projectClass/contrastsClass.m
```diff
@@ -71,7 +71,7 @@
             contrastBackgroundActions = cell(1,nContrasts);
             contrastBulkIns = ones(1,nContrasts);
             contrastBulkOuts = ones(1,nContrasts);
-            contrastDomainRatios = zeros(1,nContrasts);
+            contrastDomainRatios = ones(1,nContrasts);
```
StephenNneji, line 74: I am not sure why this is changed?

targetFunctions/+domainsTF/customXY.m
```diff
@@ -13,11 +13,19 @@
     % Extract individual parameters from problemStruct
     [numberOfContrasts, ~, contrastBackgroundIndices, contrastQzshiftIndices,...
      contrastScalefactorIndices, contrastBulkInIndices, contrastBulkOutIndices,...
-     contrastResolutionIndices, contrastDomainRatioIndices, backgroundParamArray,...
+     contrastResolutionParamIndices, contrastDomainRatioIndices, backgroundParamArray,...
      qzshiftArray, scalefactorArray, bulkInArray, bulkOutArray, resolutionParamArray,...
      domainRatioArray, dataPresent, nParams, params, ~, ~, contrastBackgroundActions,...
      cCustFiles, useImaginary] = extractProblemParams(problemStruct);
 
+    numDomainRatios = length(domainRatioArray);
```
StephenNneji, line 21: Is this missed or can it not be moved?

targetFunctions/+domainsTF/standardLayers.m
```diff
@@ -20,11 +20,19 @@
...
     [numberOfContrasts, geometry, contrastBackgroundIndices, contrastQzshiftIndices,...
      contrastScalefactorIndices, contrastBulkInIndices, contrastBulkOutIndices,...
-     contrastResolutionIndices, contrastDomainRatioIndices, backgroundParamArray,...
+     contrastResolutionParamIndices, contrastDomainRatioIndices, backgroundParamArray,...
      qzshiftArray, scalefactorArray, bulkInArray, bulkOutArray, resolutionParamArray,...
      domainRatioArray, dataPresent, nParams, params, ~, resample,...
      contrastBackgroundActions, ~, useImaginary] = extractProblemParams(problemStruct);
-    
+
+    numDomainRatios = length(domainRatioArray);
```
StephenNneji, line 28: Also here, is this missed or can it not be moved?

targetFunctions/common/checkIndices.m
```diff
@@ -0,0 +1,60 @@
+function checkIndices(problemStruct)
```
StephenNneji, line 1: This is a duplicate of _API/checkIndicies.m_

API/projectClass/contrastsClass.m
DrPaulSharp: Just for consistency with the rest of the arrays and tests now that each entry is filled in or set to -1

targetFunctions/+domainsTF/customXY.m
DrPaulSharp: Good job someone is paying attention . . .

targetFunctions/common/checkIndices.m
DrPaulSharp: Yeah, I moved it and forgot to remove it from the repo.

API/projectClass/contrastsClass.m
DrPaulSharp: A few lines further down, we loop through each contrast and set the value if the contrast has a domain ratio, or set it to -1. I therefore can't see any issue with this change. Having said that, I take your point that there is potential for problems with this default, so I'll change back. I'm inclined to think, especially with the check indices routine now present, that setting all of these to a default of zero is the better option. One to talk over on Monday maybe?

StephenNneji (approved):

Issue #180, opened 2023-11-29
Implement copy functionality for projectClass?


arwelHughes:
I've noticed that trying to copy projectClass makes a new reference to the existing object, rather than a copy of it. This is the default behaviour for handle classes, but can be overridden..

https://uk.mathworks.com/help/matlab/matlab_oop/custom-copy-behavior.html

It would be useful to implement a copy method for projectClass()

DrPaulSharp:
Whilst the implementation above does enable a handle class to be copied, it only allows for a shallow copy - i.e., references in the object are copied, and hence point to the same object. This means that for projectClass, we can make a copy of the object, and changing (say) the experiment name behaves as expected:

![image](https://github.com/RascalSoftware/RAT/assets/44529197/cb3b6812-bac9-4da6-bb94-6ccd1bb0a4d7)

but when adding a layer, the reference for both p and q is to the same layersClass object, so we have the following.

![image](https://github.com/RascalSoftware/RAT/assets/44529197/f4296c96-8855-4cfe-ab05-31ed953945e1)

Matlab does not seem to have the capacity to make deep copies that would resolve this issue: https://uk.mathworks.com/matlabcentral/answers/41674-deep-copy-of-handle-object

PR #239, opened 2024-06-20
Adds clone function for project class

StephenNneji:
A hack to solve #180. Clone is technically not equal but should be good enough since we use writescript


tests/testExamples.m
```diff
@@ -139,7 +139,9 @@ function testWriteScript(testCase, exampleScriptFile)
             close all;
 
             % Write the script
-            problem.writeScript(objName="scriptProblem", script="projectScript");
+            clonedProblem = problem.clone();
+            testCase.verifyNotEqual(clonedProblem, problem);
```
DrPaulSharp, line 143: Given that this test does not look intuitively correct, I think it might be worth adding a comment stating that this line tests that clone produces a deep copy of the project.

DrPaulSharp (approved):

PR #242, opened 2024-07-03
Custom file Changes

StephenNneji:
* Custom files will now accept filename without extension if no extension is given and the language is python or matlab the py or m extension will be appended to filename
* Test folders were removed from the addpaths list to avoid these files conflicting with any others on path 
* Modifies clone function so data file is not written to disk
* Refactors test to use PathFixture to add paths for test data


API/projectClass/customFileClass.m
```diff
@@ -43,8 +43,8 @@ function delete(obj)
             % A custom file entry can be added with no parameters, just the
             % name of the custom file entry, the name of the entry
             % alongside a filename, or can be fully defined by specifying
-            % the name of the custom file entry, filename, language, and
-            % file path. For MATLAB, the provided path must be in the
+            % the name of the custom file entry, filename, language, 
+            % file path, and. For MATLAB, the provided path must be in the
```
DrPaulSharp, line 47: I'm guessing this isn't right . . . Is the old version correct?

DrPaulSharp: ahh - "and function name"?

API/projectClass/customFileClass.m
```diff
@@ -319,9 +320,31 @@ function displayTable(obj)
...
+                return
+            end
+            
+            [~, ~, ext] = fileparts(filename);
+            if isempty(ext)
+                if strcmp(language, supportedLanguages.Matlab.value)
+                    filename = [filename, '.m'];
+                elseif strcmp(language, supportedLanguages.Python.value)
+                    filename = [filename, '.py'];
+                end
```
DrPaulSharp, line 339: Do we want to raise a warning/error if language is C++ and extension is empty?

DrPaulSharp (approved):

API/projectClass/customFileClass.m
StephenNneji: Dylibs are binary files so extension doesn't matter, I will leave as is for now

PR #250, opened 2024-07-30
Fixes a typo in `projectClassToR1`

alexhroom:
The `projectClassToR1` function misspells the R1 struct field `resolution_constr` as `reslution_constr`, meaning that the field `resolution_constr` is missing from the resulting R1 struct. This PR fixes that. (for some reason the test data contained both `resolution_constr` and `reslution_constr`, the latter of which has been removed from all test data - i assume `projectClassToR1` was used to generate the test data?)


DrPaulSharp:
Yes - conversion to and from R1 files comes up in both the test data and also the examples. As you've got the tests to pass I'm pretty confident you've caught it everywhere.

DrPaulSharp (approved):

PR #251, opened 2024-07-31
Add contrast names to plot event data

StephenNneji:


DrPaulSharp (approved):
Thanks for doing this Stephen, it's a lot more code than it probably should have been!

PR #255, opened 2024-08-05
Adds a stop event that works via a temp IPC file

StephenNneji:
- Adds initial design for IPC for the stop event
- Ensure DREAM and NS respect the display off option
- Unifies spelling of initialise in the repo
- Refactors text progress bar to respect display off

Example usage 

```
% Send everything to RAT....
controls = controlsClass();
controls.procedure = 'ns';

% initialise inter-process communication
controls.initialiseIPC();

% Send everything to RAT....
pool = parpool("Processes", 1);
f = parfeval(pool, @(varargin) evalc('RAT(varargin{:})'), 3, problem, controls);
pause(20)
controls.sendStopEvent();
[output, problem, results] = fetchOutputs(f);
disp(output)

plotRefSLD(problem, results);
delete(gcp('nocreate'))
```


StephenNneji:
I ran on IDAaas with `timeit` using the `standardLayersDSPCScript` with the following controls. Seems decent to me, what are your thoughts @arwelHughes 

```
controls = controlsClass();
controls.procedure = 'simplex';
controls.maxIterations = 100000;
controls.maxFuncEvals = 100000;
```

Original: 7.3198sec
IPC: 9.3706sec

```
controls = controlsClass();
controls.procedure = 'dream';
controls.nSamples = 1000000;
```

Original: 373.1561sec
IPC: 401.5368sec

DrPaulSharp (approved):

arwelHughes:
That's nearly 30% slower for example 1.... not very good.
I will never be comfortable having file handling at every iteration tbh...
Since we can use C++ I would be much happier with shared memory IPC tbh...

StephenNneji:
@arwelHughes Did you look at the number of iterations/samples in both examples?  I don't think anyone will be run these numbers and even though they do the hit is a few seconds on IDAaas which is very likely the worst case 
Also note that this hit is only present when IPC is enabled in the controls class, it is disabled by default

arwelHughes:
"it is disabled by default...."
Enabled by default in the GUI no?

I'm just conceptually allergic to files, really! 🙂

StephenNneji:
> Enabled by default in the GUI no?

@arwelHughes, Yes it will be enabled for the GUI.

> I'm just conceptually allergic to files, really!

I am not a big fan myself 😄 but this is the solution that gets us moving forward and as I mention I will revisit when I get a better idea.

PR #256, opened 2024-08-07
Adds code to check for invalid procedures in RATMain

DrPaulSharp:
This code is added to warn of the bug scenario we encountered last week of an invalid procedure making it into the compiled RATMain.

### DIFF
--- a/API/RATMain.m
+++ b/API/RATMain.m
@@ -35,8 +35,7 @@ switch controls.procedure
         end
         [problemStruct,result,bayesResults] = runDREAM(problemStruct,problemCells,problemLimits,controls,priors);
     otherwise
-        throw(exceptions.invalidOption(sprintf('procedure must be a procedures enum or one of the following strings (%s)', ...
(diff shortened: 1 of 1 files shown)


API/RATMain.m
```diff
@@ -34,6 +34,8 @@
             fprintf('\nRunning DREAM\n\n');
         end
         [problemStruct,result,bayesResults] = runDREAM(problemStruct,problemCells,problemLimits,controls,priors);
+    otherwise
```
StephenNneji, line 37: This should work for enums

`error('The procedure "%s" is not supported. The procedure must be one of "%s"', controls.procedure, strjoin(fieldnames(coderEnums.procedures), '", "'));`

StephenNneji (approved):

API/RATMain.m
DrPaulSharp: I was trying that and having the build fail. I'll give this a go and see if it works.

DrPaulSharp: That seems to build, so we're all good.

PR #257, opened 2024-08-13
Refactors to improve performance

DrPaulSharp:


API/makeEmptyResultStruct.m
```diff
@@ -35,19 +35,19 @@
     % (2) result.contrastParams
 
     backgroundParams = zeros(nContrasts,1);
-    coder.varsize('backgroundParams',[10000 1],[1 0]);
+    coder.varsize('backgroundParams',[100 1],[1 0]);
```
StephenNneji, line 38: Where are these new values from?

StephenNneji (commented):
Hey Paul, please this PR need clarification. I can kind of see the attempt to improve memory by changing the array sizes on the input array but I am not sure where the new values come from and I think there is now a mismatch between the input values in `makeCompileArgs` and the result values you've set.

I don't fully understand the reasons behind the refactors to `callReflectivity` and if or how they improve performance

API/makeEmptyResultStruct.m
DrPaulSharp: They're arbitrary, as with the previous values.

We've seen that the code uses inordinate amounts of memory so I'm proposing a limit of 100 values per data block (i.e, for each of params, background_params, scalefactors etc). This would correspond to 33 independent layers, which is far more than any of our examples use, though I accept there may be cases where this limit is too low.

targetFunctions/common/callReflectivity/callReflectivity.m
```diff
@@ -26,25 +26,19 @@
 % Make arrays for thick, sld, rough
 thicks = zeros(nLayersTot,1);
 sldArray = zeros(nLayersTot,1);
+slds = complex(sldArray,sldArray);
 roughs = zeros(nLayersTot,1);
 
-if useImaginary
```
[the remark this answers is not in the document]
DrPaulSharp: This needs to be read in conjunction with the `abelesParallelPoints` and `abelesSingle` routines. In those, if the `slds` array (referred to there as `layers_rho`) is real, it is made into a complex array. So either we start with a complex array if `useImaginary=true` or we start with a real array and later made it complex if `useImaginary=false`. However, in the latter case, the checking if the array is real is done using an array lookup and if statement within a (potentially parallel) loop. Hence, I've removed this check and made sure the array is always complex from the beginning.

I do want to check about the use of "eps" and "tiny" - it'd be good to have a consensus on those in the code.

DrPaulSharp:
Good point about the limit in `makeCompileArgs` - there's not much point limiting the values on the way out but not the way in! I'm hopeful that reducing the limit will improve the memory issues we have encountered, and it's probably best now to grasp the nettle and make sure we do it in all places. The code does build correctly, but leave it with me and I'll see if we find any improvement in memory/performance as the limit is reduced.

targetFunctions/common/callReflectivity/callReflectivity.m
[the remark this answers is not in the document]
DrPaulSharp: A very quick check showed a reduction of 0.3s on a 10s calculation in serial. Hardly earth-shattering but I'll look for a more illustrative example, there are algorithms/settings that run many reflectivity calculations which might show a better improvement, and see if parallel points gives a good indication too.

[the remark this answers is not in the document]
DrPaulSharp: Running a NS calculation on the non polarised custom XY example parallelised over points takes an average 120.837s on the main branch, and 105.722s on this branch.

compile/fullCompile/makeCompileArgsFull.m
```diff
@@ -3,8 +3,9 @@
 % Define the arguments for compiling reflectivityCalculation
 % using codegen.
 
-%% Define argument types for entry-point 'reflectivityCalculation'.
+%% Define argument types for entry-point 'RATMain'.
 maxArraySize = 10000;
+maxDataSize = 10000;
```
StephenNneji, line 8: I don't think these bounds are needed. Theses values should just be `Inf `except where we actually have a bound, to the best of my knowledge the memory is allocated as needed so not sure the reason for the artificial handicap. @arwelHughes do you remember the original reason for the max size?  

Is it possible that something like the data (`problemCells{2} = inputStruct.data`) can exceed our artificial limit of 10000 rows maybe someone doing a simulation or something?

minimisers/generalUtils/makeEmptyBayesResultsStruct.m
```diff
@@ -47,23 +47,23 @@
...
     for i = 1:nContrasts
         reflectivityXData{i} = xDataCell;
     end
     
     if isDomains
         sldXData = cell(nContrasts,2);
-        sldXDataCell = [1 1 1; 1 1 1];
-        coder.varsize('sldXDataCell',[2 1e4],[1 1]); 
+        domainSldXDataCell = [1 1 1; 1 1 1];
+        coder.varsize('domainSldXDataCell',[2 1e4],[0 1]);
```
StephenNneji, line 58: I think this is meant to be the other way around

`coder.varsize('domainSldXDataCell',[2 1e4],[1 0]); `

DrPaulSharp: Are you sure? I've written that the fixed dimension has size 2 since there are always two domains, and the variable size dimension has size up to 1e4: https://uk.mathworks.com/help/coder/ref/coder.varsize.html

arwelHughes:
I think the memory allocation ‘under the hood’ is different. To quote from the TMW site….


·         “If you do not specify upper bounds with a coder.varsize declaration and the code generator is unable to determine the upper bounds, the generated code uses dynamic memory allocation. Dynamic memory allocation can reduce the speed of generated code. To avoid dynamic memory allocation, specify the upper bounds by providing the ubounds<https://uk.mathworks.com/help/coder/ref/coder.varsize.html#mw_f8911ef0-febf-4349-b494-360c5163671e> argument.
”

Although I don’t know the exact details, a fixed upper bound always allocates in stack, using Inf for the upper bound enforces DMA, so that is in heap, which is apparently slower. My understanding is that an Inf upper bound has some performance hit, regardless of the eventual size of the array…

(e.g. see https://uk.mathworks.com/matlabcentral/answers/500451-why-do-i-always-need-coder-varsize-directives )

From: StephenNneji ***@***.***>
Date: Tuesday, 20 August 2024 at 12:42
To: RascalSoftware/RAT ***@***.***>
Cc: Hughes, Arwel (STFC,RAL,ISIS) ***@***.***>, Mention ***@***.***>
Subject: Re: [RascalSoftware/RAT] Refactors to improve performance (PR #257)

@StephenNneji commented on this pull request.

In compile/fullCompile/makeCompileArgsFull.m<https://github.com/RascalSoftware/RAT/pull/257#discussion_r1723121249>:

>  maxArraySize = 10000;

+maxDataSize = 10000;

I don't think these bounds are needed. Theses values should just be Inf except where we actually have a bound, to the best of my knowledge the memory is allocated as needed so not sure the reason for the artificial handicap. @arwelHughes<https://github.com/arwelHughes> do you remember the original reason for the max size?

Is it possible that something like the data (problemCells{2} = inputStruct.data) can exceed our artificial limit of 10000 rows maybe someone doing a simulation or something?

In minimisers/generalUtils/makeEmptyBayesResultsStruct.m<https://github.com/RascalSoftware/RAT/pull/257#discussion_r1723149416>:

>      for i = 1:nContrasts

         reflectivityXData{i} = xDataCell;

     end



     if isDomains

         sldXData = cell(nContrasts,2);

-        sldXDataCell = [1 1 1; 1 1 1];

-        coder.varsize('sldXDataCell',[2 1e4],[1 1]);

+        domainSldXDataCell = [1 1 1; 1 1 1];

+        coder.varsize('domainSldXDataCell',[2 1e4],[0 1]);

I think this is meant to be the other way around

coder.varsize('domainSldXDataCell',[2 1e4],[1 0]);

StephenNneji:
@arwelHughes The link you shared says

> Whenever the upper bounds are fixed variables can be still allocated in stack, so the dynamic memory alloaction may not be required (not always true !) .

Stack allocation is not guaranteed which makes sense as the stack can be as small as 1MB on some machines. From my look at the generated code we are already using the heap a lot even when we set an upper bound, it seems as longer as both dimension are not fixed it used the heap array so for example

`coder.varsize('outlierChains',[1e3 1e3],[1 1]);` used a bounded array on stack because both dimensions are fixed whereas `coder.varsize('percentile65',[2 1e3],[0 1])` uses heap memory so we are not getting any benefit from setting the upper bound on variable sized arrays

API/makeEmptyResultStruct.m
```diff
@@ -61,84 +63,78 @@
...
     else
         resampledLayers = cell(nContrasts,1);
-        resampledLayersCell = [1 1 1; 1 1 1];
-        coder.varsize('resampledLayersCell',[10000 3],[1 0]);
         for i = 1:nContrasts
             resampledLayers{i} = resampledLayersCell;
         end
     end
     
     fitParams = zeros(1,nParams);
```
StephenNneji, line 132: Should this match the other fitParams that is now a row array?

StephenNneji (approved):

API/makeEmptyResultStruct.m
DrPaulSharp: Seeing as it's a simple change, yes!

PR #263, opened 2024-09-02
Splits `resampleParams` into two params

alexhroom:
see RascalSoftware/python-RAT#69 - this PR splits `resampleParams` into two parameters, `resampleMinAngle` and `resampleNPoints`.


DrPaulSharp (approved):
All looks good, a few comments to consider.

PR #272, opened 2024-10-08
Fix DE crash

### LINKED ISSUE #271: Running RAT with default project class and DE controls leads to crash
The following crashes in MATLAB and Python 
```
>> p = projectClass();
>> c = controlsClass();
>> c.procedure = 'de';
>> RAT(p, c)
```

The error message is 

```

Error using parametersClass/setValue
Parameter value 0.000000 must be within the limits 1.000000 -- 5.000000

Error in projectClass/setParameterValue (line 346)
            obj.parameters.setValue(row, value);

Error in parseOutToProjectClass (line 9)
    project.setParameterValue(i,params(i));

Error in RAT (line 30)
project = parseOutToProjectClass(project,problemStruct);
```

StephenNneji:
Fixes #271 

The problem was caused by the DE initializing the best results to zeros but the while loop never runs so the result remained zero  

https://github.com/RascalSoftware/RAT/blob/e2067953e504dd81bb0f7f5a7847d9ee419f4850/minimisers/DE/deopt.m#L182-L191

The fix resets the best results to the initial values if the while loop will not be run.

### DIFF
--- a/API/projectClass/parametersClass.m
+++ b/API/projectClass/parametersClass.m
@@ -108,7 +108,7 @@ classdef parametersClass < tableUtilities
                 
                 % Type validation
                 if ~isnumeric(values)
-                    throw(exceptions.invalidType('Expecting numeric values as params 2 - 4'));
+                    throw(exceptions.invalidType('Lower limit, value, and upper limit (2nd, 3rd and 4th arguments) must be numeric values'));
                 end
                 
                 if values(1) > values(3)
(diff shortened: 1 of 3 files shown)


StephenNneji:
```
>> p = projectClass();
>> c = controlsClass();
>> c.procedure = 'dream';
>> RAT(p, c)
```
Dream was also crashing this is now fixed

DrPaulSharp (approved):

PR #276, opened 2024-10-23
Add Data and Function Background types

DrPaulSharp:
This enables data and function background types. In the compiled code, we construct a background array over the same range of q points that is used for the simulation, which replaces `backgroundParams` in the output.

Constant backgrounds are now a constant array made in `constructBackground`.

Data backgrounds are defined in the data block of the project and appended to the contrast data in `insertDataBackgroundIntoContrastData` (called in `parseClassToStructs`), with an optional offset added in `constructBackground`.

Function backgrounds are defined in the custom files block of the project and evaluated and applied in `constructBackground`. Note that only MATLAB background functions are supported at present.

The API for backgrounds and resolutions has changed. We have added the field "source" which contains either the background/resolution parameter, entry in the data block, or entry in the custom file block for constant, data, and function backgrounds respectively. This differentiates this field from the five "value" fields, which now ONLY contain background parameters (the optional offset for data, and up to five function parameters).
NOTE - I'm not entirely sure about the name "source" for this field, feedback on something better is welcome. I've kept it at five value fields, meaning a function background can have up to five arguments, which feels sensible to me.


DrPaulSharp (commented):
Tests are required for the new routine "constructBackground". At present our examples and test data consider only constant backgrounds, so our data to test the function section here is limited. How should we approach testing for this routine?

API/projectClass/backgroundsClass.m
```diff
@@ -125,7 +134,7 @@
             % Changes the value of a given background in the table. Expects the 
             % index or name of background and keyword/value pairs to set. 
             %
-            % background.setBackground(1, 'name', 'back 1', 'type', 'constant', 'value1', 'param_name');
+            % background.setBackground(1, 'name', 'back 1', 'type', 'constant', 'source', 'param_name');
             if isText(row)
```
StephenNneji, line 138: I think this function  `setBackground` needs to be refactored  because it is possible to change the type without updating the source also when I change the **source**, **value 1** to **value 5** remains the same which can have unexpected consequences for example if I switch from **function** to **data**, **value 1** now becomes the unwilling offset for my data and this will not be caught by any checks

StephenNneji (commented):
Looks good please see comments. I think its fine if `constructBackground` has no test as its being called by the example tests but if you are up for it we could save the inputs and output that goes in to the function from the two examples and have its own independent test. What I would like in the future is some kind of validation (maybe from ORSO) for the background stuff

API/projectClass/backgroundsClass.m
```diff
@@ -79,37 +80,45 @@
...
+                       % the custom files table
+                       newRow{3} = in{3};
+
+                       if length(in) >= 4
+                           % Any other given parameters must be valid
+                           % background parameters
+                           params = in(4:end);
+                           params = params(~(cellfun(@(x) isequal(x,""), params)));
+                           for i = 1:length(params)
+                              thisParam = obj.validateParam(params(i));
```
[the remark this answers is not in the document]
DrPaulSharp: Hmm, that sounds like a recipe for spaghetti code to me. Because of the unique way the matlab API of RAT is written, we have no idea if the data/custom file entry is even defined when the background is defined. Ultimately, this is why background parameters and backgrounds are defined in the same class. Running this check in `parseClassToStructs` feels the only sensible way to proceed to me.

API/projectClass/backgroundsClass.m
DrPaulSharp: I see two options here: either bar the user from changing the background type, or have the result of changing the type being that the source and all values are cleared. I think the second of these is not intuitive to users, so we should take the first option and make sure the background type is not changable. This will also be easy to transfer to the python, since changing the type there without updating anything else will trigger validation errors in the source field.

API/projectClass/backgroundsClass.m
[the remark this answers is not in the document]
StephenNneji: I am just concerned about the inconsistency where we validate params on setting them but don't validate the data or custom files. I don't think holding references to the files and data objects and checking names will lead to spaghetti code except I am missing something. Note background params are not defined in the class they are passed in to the class on initialization https://github.com/RascalSoftware/RAT/blob/fcc187dd175291638c3628cc7a30605fa39f6078/API/projectClass/backgroundsClass.m#L30-L38

API/projectClass/backgroundsClass.m
StephenNneji: I think barring the user from changing the type is the confusing option. You mean for type alone the user will have to delete the background and add a new one. I would prefer the source and values1-5 are reset if the user changes the type without setting them and a warning is printed to notify the user what and why

DrPaulSharp: I see what you mean, but in ALL cases of changing the type, the source field must also change. Changing the type and the source is not far off defining a new background anyway. Also, the old background does not have to be deleted, it can be retained and switched in the contrast(s). I can see the user case of having a data background and a function background and wanting to try out each one.

API/projectClass/backgroundsClass.m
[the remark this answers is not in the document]
StephenNneji: I don't see the fundamental difference these objects are referred to by the backgrounds so background holds a reference if you are concerned about access to the reference they could be made private. Passing allNames to the add and set function seems like a round about way to do the same thing

API/projectClass/backgroundsClass.m
StephenNneji: My concern as always is consistency the set function should set stuff not direct the user to the add function to add a new background even if the user has to set most of the fields. If the user wants to alternate between types they will use the add function instead.

API/projectClass/backgroundsClass.m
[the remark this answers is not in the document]
DrPaulSharp: The point I'm making is that `project.backgrounds.backgroundParams` is defined, but `project.backgrounds.data` isn't.

API/projectClass/backgroundsClass.m
DrPaulSharp: I get your point, but setting one field and clearing six others isn't great for consistency either. Hmmm.

StephenNneji: So what is your solution for telling the user they cannot change the type  field? throw an exception?

DrPaulSharp: Yeah, I think so. We raise an exception for trying to remove a protected parameter, so it falls in that area I guess. I'll have to have a think about that one.

StephenNneji: We do not throw any exception anywhere for changing a field, we throw a warning and reset the data range, we throw a warning an reset the prior mu and sigma. The reason for not deleting a protected parameter is well understood by the user

DrPaulSharp: If you change the min/max/value of a parameter range such that: "NOT (min<=value<=max)", an exception is thrown.

API/projectClass/backgroundsClass.m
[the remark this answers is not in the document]
StephenNneji: I am not sure how that affect the decision to hold a reference to the 2 objects. The references could be private so no one can access it from the background object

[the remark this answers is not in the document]
DrPaulSharp: Right, I'll look into this further. But the point is that data and custom files cannot be handled identically to background parameters, because they are separate data blocks defined outside of the scope of the background object.

examples/miscellaneous/roundRobin/roundRobin.mlx
StephenNneji, line 1: Did you intend to change this?

DrPaulSharp: Yes, I needed to change "value1" to "source" for the backgrounds fields.

API/projectClass/backgroundsClass.m
```diff
@@ -121,11 +138,11 @@
             obj.backgrounds.removeRow(row);
         end
         
-        function obj = setBackground(obj, row, varargin)
+        function obj = setBackground(obj, row, allowedNames, varargin)
```
StephenNneji, line 141: Is it possible to refactor so the setters are called after all the validation is done? This way if any validation fails you have not changed half of the parameters already. 
```
% All the Validation(
...
 obj.backgrounds.setValue(row, 1, inputBlock.name);           
 obj.backgrounds.setValue(row, 2, inputBlock.type);
 obj.backgrounds.setValue(row, i + 3, value);
```

API/projectClass/backgroundsClass.m
```diff
@@ -152,16 +171,71 @@
...
             end
-            values = {inputBlock.value1, inputBlock.value2, inputBlock.value3, inputBlock.value4};
-            for i = 1:4
+
+            % For data and function types, source is the data/function name
+            source = convertStringsToChars(inputBlock.source);
+            if ~isempty(source)
+                if strcmpi(inputBlock.type, allowedTypes.Constant.value)
+                    source = obj.validateParam(source, obj.backgroundParams.getNames(), 'Background Param');
+                    maxValues = 0;
```
StephenNneji, line 207: `maxValues` is used in multiple places might be worth making a private/constant struct with these values are reusing that

API/projectClass/backgroundsClass.m
```diff
@@ -204,33 +278,32 @@ function displayBackgroundsObject(obj, showPriors)
...
     end
 
-    methods (Access = protected)
-        function thisPar = validateParam(obj, param)
+    methods (Static, Access = protected)
+        function thisPar = validateParam(param, paramList, parameterType)
             % Checks that given parameter index or name is valid, then returns the
             % parameter name. 
             %
             % param = obj.validateParam('param_name');
```
StephenNneji, line 294: please update comments

StephenNneji (commented):
Looks good please see comments, in addition to previous comments about the resolutionsClass

StephenNneji (approved):
Looks good, Thanks for persevering on this one.

API/projectClass/customFileClass.m
```diff
@@ -238,6 +238,7 @@ function displayTable(obj)
             %
             % customFiles.toStruct()
             fileStruct.files = {};
+            fileStruct.fileIdentifiers = {};
```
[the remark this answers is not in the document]
DrPaulSharp: If we just take the first column of an empty table it's an empty string array, rather than an empty cell array, so I'll leave that as it is.

PR #277, opened 2024-11-01
Reduces "bayesShadedPlot" to use only the "mean" fit option

### LINKED ISSUE #275: Decide on "mean" or "average" reflectivity and SLD for Bayes Shaded plots
For Bayes shaded plots there is the option to plot what is described as the "mean" (reflectivity and SLD are those from the calculation using the best parameter values) and the "average" (reflectivity and SLD values from calculations of a range of parameter values are averaged).

We need to decide which of these should be used and implement it consistently, and remove any unnecessary data.

#249 begins this work.

DrPaulSharp:
Closes #275


StephenNneji (approved):
Looks good please see comment

Issue #299, opened 2024-12-10
Best fit parameters not present in results struct 


maxskoda:
* **This issue is a** [bug report]

* **What is the current behaviour?**
After performing a fit or a DREAM run, the results.fitParams struct is just a single value.

* **If the current behaviour is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** 
..just run a fit and check contents of results.fitParams

* **What is the expected behaviour?**
results.fitParams should be an array containing the final best fit parameters.

* **Please tell about the environment(s) affected:**
  
  - Version: [e.g. 1.0.0-beta]
  - Operating System: [e.g. All | Windows (10, 8, 7) | Linux (Ubuntu, Fedora, RedHat etc)]
  - Platform: [e.g. Name processor and graphics card | IDAaas | Remote Desktop Connection]

  
* **Other information (e.g. detailed explanation, screenshots, stack traces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc):**

DrPaulSharp:
Based on what we've found with this bug, we also need to consider how we test the translation of python to C++ and back.

PR #301, opened 2024-12-16
Converts "fitParams" and "otherParams" to row arrays

DrPaulSharp:
This PR addresses the issue in #299 

The previous PR #257 unified the code to use column arrays for `problemStruct.fitParams` following the lead of the `packParams` routine. However, this bug has arisen because the de and dream algorithms expect `fitParams` to take row arrays. However, `simplex` expects `fitParams` to take column arrays.

The best and least intrusive solution to this is to refactor the code, including `packParams`, to use row arrays, with two changes to `simplex` to enable it to work correctly.


DrPaulSharp (commented):
Based on what we've found with this bug, we also need to consider how we test the translation of python to C++ and back, and we need a basic test of "simplex", "de", "ns" and "dream" calculations in the python too - it seems like this bug exposed a seg fault there that has gone unnoticed until now.

compile/fullCompile/makeCompileArgsFull.m
```diff
@@ -56,8 +56,8 @@
 ARGS_1_1.numberOfDomainContrasts = coder.typeof(0);
 ARG = coder.typeof(0,[1 maxArraySize],[1 1]);
 ARGS_1_1.domainContrastLayers = coder.typeof({ARG}, [1 maxArraySize],[0 1]);
-ARGS_1_1.fitParams = coder.typeof(0,[maxArraySize 1],[1 0]);
-ARGS_1_1.otherParams = coder.typeof(0,[maxArraySize 1],[1 0]);
+ARGS_1_1.fitParams = coder.typeof(0,[1 maxArraySize],[0 1]);
```
StephenNneji, line 59: The PR description said "refactor the code, including packParams, to use row arrays" but this is a column array ?

StephenNneji (commented):
I agree with the description that changing the simplex to used row arrays should be the easiest route to fix this problem but this PR changes everything to column arrays. Did something change the described direction?

compile/fullCompile/makeCompileArgsFull.m
DrPaulSharp: Ok, my understanding is this array has a single row, and multiple columns, something like [0, 0, 0]. Does that match what you think?

DrPaulSharp:
Yeah, I did have to switch tack on this one to keep things as simple as possible. I'll just go through and verify everything and then update as necessary.

compile/fullCompile/makeCompileArgsFull.m
StephenNneji: You are right, mix up in head

DrPaulSharp:
Ok, this PR changes everything to row arrays in the code outside of the minimisers. There are changes to simplex to ensure that the column arrays it uses are transposed to row arrays outside of the minimiser. There are also changes to the other minimisers to make sure that they convert (or not) to row arrays outside of the minimiser, rather than working with columns outside the minimiser as was previously done. Does that cover everything you wanted to know?

StephenNneji (approved):
looks good, sorry for the confusion, I'm probably just recovering from my break :smile:

PR #303, opened 2025-01-07
Nested Sampler Fix

arwelHughes:
Added the fix to the Nested Sampler to avoid the fractional matrix power.

Removed the 'roundRobin' example....

Replaced it with a direct model selection example using NS ('d2oCheck')

Modified the test examples to remove the roundRobin test, and replaced it with the new example.


minimisers/NS/calcEllipsoid.m
```diff
@@ -65,7 +65,17 @@
...
 fV = 1;
 if VE < VS
-    fV = (VS/VE)^(2/ndims);
+
+    
+    % Fix to avoid the fractional matrix power error....
+    % Replace....
+    %fV = (VS/VE)^(2/ndims);
+
+    % .... with....
```
alexhroom, line 74:
```suggestion
    % the original implementation calculates this as
    %fV = (VS/VE)^(2/ndims);
    % however when compiled to C++,
    % MATLAB Coder does not compile the code for fractional powers of matrices.
    % so we must replace it with the explicit calculation:
```

alexhroom, line 74:
```suggestion
    % the original implementation calculates this as
    %fV = (VS/VE)^(2/ndims);
    % however when compiled to C++,
    % MATLAB Coder does not compile the code for fractional powers of matrices.
    % so we must replace it with the explicit calculation:
```
just to make this comment more verbose, because "the fractional matrix power error" means nothing to anyone who wasn't in that meeting!

alexhroom:
I can't seem to reproduce the bug on main... what was it again? Just a random crash sometimes when NS runs?

StephenNneji:
> I can't seem to reproduce the bug on main... what was it again? Just a random crash sometimes when NS runs?
 
Sorry, I was not aware this was fixing a crash, this might have been when I was away

alexhroom:
@StephenNneji what I think was happening was that in this block of calcEllipsoid:
```matlab
    % the original implementation calculates this as
    %fV = (VS/VE)^(2/ndims);
    % however when compiled to C++,
    % MATLAB Coder does not compile the code for fractional powers of matrices.
    % so we must replace it with the explicit calculation:
    x = 2/ndims;
    A = log(VS/VE) * x;
    fV = exp(A);
```
as the code says, when it had `fV = (VS/VE)^(2/ndims);`, the compiled code would only work if `2/ndims` was an integer, and otherwise it would raise an error

alexhroom:
@arwelHughes do you have any scripts which cause the error that we've fixed here?

alexhroom (approved):

PR #311, opened 2025-01-13
Fixes exception build issue

StephenNneji:
This should fix 

> error: no matching function for call to ‘std::exception::exception(const char*&)’


alexhroom (approved):
builds successfully on my machine :)

PR #314, opened 2025-01-14
Non-fit parameters from RasCAL-1 projects now have their constraints adjusted

### LINKED ISSUE #291: Allow invalid non-fit parameters in RasCAL-1 conversion
Apparently, it was quite popular among RasCAL-1 users to just give a value and not a minimum or maximum for a parameter that was not being fit (as the minimum/maximum don't matter there). This means that there may be many RasCAL-1 projects with invalid parameter values (i.e. value larger than min or smaller than max). `r1ToProjectClass` should allow these parameters, fix the minimum and maximum values, and provide a warning that it has done so.

alexhroom:
Fixes #291. If a RasCAL-1 project has a non-fit parameter with `min > value` or `max < value`, as was allowed in RasCAL-1, this is adjusted and the user is warned.


DrPaulSharp (commented):
The function solving this problem is good, alongside the comment below, I think you'll need to apply this to the standard params block (i.e., `problem.params`), and the substrate roughness.

DrPaulSharp (approved):

PR #318, opened 2025-01-15
Adds MATLAB and Cpp custom function support for backgrounds

### LINKED ISSUE #287: Implement C++ and Python custom functions for the backgrounds

StephenNneji:
Fixes #287


DrPaulSharp (approved):
All looks good. Have you tested with a python background function? I'll put in a PR to convert the background function in the python API to be written in python instead of matlab.

PR #327, opened 2025-01-22
Adds `plotChain` convenience function and fixes `plotHists`

### LINKED ISSUE #321: Make chain plots more user friendly
Currently, the MATLAB chain plot invocation is
```matlab
mcmcplot(results.chain,[],results.fitNames,'chainpanel');
```

we should simplify this to a `plotChain` function which just takes `results` as an argument.

### LINKED ISSUE #328: plotHists expects a figure
`plotHists` expects a figure  `f`

```
function h = plotHists(results,f,varargin)   
    % Plots the Bayes histogram plot from the chain, with or without smoothing.
    % If selected, smoothing is via a moving average algorithm.
    
    % Parse the inputs...
    % Parse the input options
    chain = results.chain;
```
which leads to incorrect behaviour for example in  `plotHists(results,'smooth',false)` 'smooth' is the the figure

Refactor the function to use an argument block and make the figure an optional argument

alexhroom:
- Adds a function `plotChain` which allows for more convenient plotting of MCMC chains. Fixes #321
- changes `plotHists` to make `figure` an optional argument. Fixes #328


utilities/plotting/plotHists.m
```diff
@@ -1,4 +1,4 @@
-function h = plotHists(results,f,varargin)   
+function h = plotHists(results,varargin)
```
StephenNneji, line 1: Please can you remove the varargin and use a argument block e.g.

```
function h = plotHists(results,options)   
    % Plots the Bayes histogram plot from the chain, with or without smoothing.
    % If selected, smoothing is via a moving average algorithm.
    arguments
        results
        options.figure {isFigure} = []
        options.smooth {logical} = true
    end
```

StephenNneji (commented):
You also need to update the examples and live scripts that use `mcmcplot` and `plotHist`

StephenNneji (approved):
looks good, thanks for the fix

PR #331, opened 2025-01-23
Changes error in customFile class to warning

StephenNneji:


DrPaulSharp (approved):

arwelHughes:
Thanks Stephen! That's an improvement.
Generally though, Matlab has it's own path control, and we seem to be trying to override it when it's not necessary.

When a file is called, it goes...


  *
Current working folder (always takes priority)..
  *
Folders on the Matlabpath list..
  *
If not found in either, it will tell you where it has found the file and ask you if you want to change directory or add the path to the pathlist.

[cid:e34b5899-9e32-4466-a50c-ef3cd6262725]

So generally, Matlab has a pretty sophisticated way of dealing with this, and we're putting another layer on top which may not be strictly necessary. I know that Python is a pain with paths (what's wrong with looking in the directory you're running in Python???), but Matlab is different. We should bear that in mind to avoid unnecessary checks in the matlab API imho.
Cheers,
A

PR #346, opened 2025-02-13
Refactors contrasts to allow model to be set with add/set methods

### LINKED ISSUE #341: validatestring as used in the contrast parser allows partial matches
In the input parser for contrasts, the routine `validatestring` is used to check the input is on the allowed lists. However, this routine will record a match if there is a partial match to the input string, e.g., `validatestring('resolution', {'resolution 1'})`.

However, in `toStruct`, which is called later, the `strcmpi` routine used there will not allow partial matches, so will fail to find the parameter in the expected list. The use of validatestring needs to be changed to allow exact matches only.

DrPaulSharp:
Closes #341 and part of #340


examples/miscellaneous/rascal1DemoProjects/monolayer7Contrast.m
StephenNneji, line 1: Is this just a format change?

API/projectClass/contrastsClass.m
```diff
@@ -234,16 +238,49 @@
                 addParameter(p,'domainRatio',   defaultDomainRatio, @(x) any(validatestring(x,expectedDomainRatio)));
```
StephenNneji, line 238: This is still using `validatestring`?

API/projectClass/contrastsClass.m
```diff
@@ -203,29 +203,33 @@
...
+            addParameter(p,'data',          defaultData,        @(x) validateExactString(x,expectedData));
 
             if obj.oilWaterCalc
                 defaultOilChiData = '';
                 addParameter(p,'oilChiData',    defaultOilChiData,  @(x) any(validatestring(x,expectedData)));
             end
 
-            addParameter(p,'background',       defaultBackground,         @(x) any(validatestring(x,expectedBackground)));
+            addParameter(p,'background',       defaultBackground,         @(x) validateExactString(x,expectedBackground));
             addParameter(p,'backgroundAction', defaultBackgroundAction,   @(x) isText(x) || isenum(x))
```
StephenNneji, line 228: missing semicolon

API/projectClass/contrastsClass.m
```diff
@@ -234,16 +238,49 @@
                 addParameter(p,'domainRatio',   defaultDomainRatio, @(x) any(validatestring(x,expectedDomainRatio)));
             end
 
+            addParameter(p,'model',            defaultModel, @(x) validateContrastModel(x,modelType,expectedModel));
+
+            % Set up the validators
+            function validateExactString(input, allowedNames)
```
StephenNneji, line 244: Why are these functions nested? The could be private with no loss of functionality

StephenNneji (commented):
Works well, please see comments

API/projectClass/contrastsClass.m
DrPaulSharp: I'll look again, but I tried having them as private and couldn't get it to work - I got an error saying it couldn't find the function.

StephenNneji (approved):
Looks good

PR #347, opened 2025-02-14
Adds remove methods for all API classes

### LINKED ISSUE #340: Make setContrast accept the model also and update remove method in project class
Add the following methods to their corresponding classes and update the calls in project class

removeData
removeCustomFile
removeLayer

DrPaulSharp:
Closes #340

### DIFF
--- a/API/projectClass/tableUtilities.m
+++ b/API/projectClass/tableUtilities.m
@@ -62,7 +62,7 @@ classdef (Abstract) tableUtilities < handle
             elseif isText(row) || iscell(row)
 
                 row = cellstr(row);
-                indices = zeros(length(row));
+                indices = zeros(1, length(row));
 
                 rowNames = obj.getNames;
                 for i = 1:length(row)
@@ -76,6 +76,7 @@ classdef (Abstract) tableUtilities < handle
 
             end
 
+            indices = sort(indices, 'descend');
             for i = 1:length(indices)
                 if indices(i) < 1 || indices(i) > obj.rowCount
                     throw(exceptions.indexOutOfRange(sprintf('Row index %d out of range 1 - %d', indices(i), obj.rowCount)));
--- a/API/projectClass/tableUtilities.m
+++ b/API/projectClass/tableUtilities.m
@@ -57,7 +57,7 @@ classdef (Abstract) tableUtilities < handle
             % Arrange parameters into a cell array
             if isa(row, 'double')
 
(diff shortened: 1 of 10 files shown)


tests/testDataClass.m
```diff
@@ -114,7 +114,27 @@ function testSetdata(testCase)
             testCase.verifyEqual(testCase.data.varTable{1, 1}, "Sim 3", 'setDataName method not working');
             testCase.verifyEqual(names.oldName, "Sim 1", 'setData returned incorrect data');
             testCase.verifyEqual(names.newName, 'Sim 3', 'setData returned incorrect data');
-         end
+        end
+
+        function testRemoveData(testCase)
+            % Checks that data can be removed by name
```
StephenNneji, line 120: Incorrect comment

API/projectClass/tableUtilities.m
```diff
@@ -44,21 +44,46 @@ function addRow(obj, varargin)
...
+
+            else
+
+                throw(exceptions.invalidType('Unrecognised Row'))
+
             end
-            
-            obj.varTable(row, :) = [];
+
+            indices = sort(indices, 'descend');
```
StephenNneji, line 79: Since the indices are sorted here, the sort on line 60 is not needed. Also the for loop below is not needed since indices is sorted.

API/projectClass/tableUtilities.m
```diff
@@ -44,21 +44,46 @@ function addRow(obj, varargin)
...
+            for i = 1:length(indices)
+                if indices(i) < 1 || indices(i) > obj.rowCount
+                    throw(exceptions.indexOutOfRange(sprintf('Row index %d out of range 1 - %d', indices(i), obj.rowCount)));
+                end
+            end
+
+            obj.varTable(indices, :) = [];
```
StephenNneji, line 86: This line should work with unsorted indices so maybe the sort can be removed

API/projectClass/tableUtilities.m
DrPaulSharp: If the input is a cell array of multiple row names, the resulting indices array will not be sorted, so it is needed here. There's also no guarantee that a list of rows provided is valid so I think the for loop is still needed too.

DrPaulSharp: Ahh, sorry. Misread here. I'll take out the sort on line 60 too.

PR #354, opened 2025-02-25
Refactors to ensure consistent use of case

StephenNneji:
This fixes some instances where `strcmp` is used instead of `strcmpi`

Also ensures that the case of a referenced parameter is always the same.  For example, currently if I add a parameter called `Oxide thick` to the parameters then add it to layers as `OXIDE THICK`, the param is added to the layers table with the exact case. Now the case should always be the same as the one in the parameter table `Oxide thick` .


API/projectClass/layersClass.m
```diff
@@ -191,18 +191,19 @@
...
+            if isText(inputVal)    
+                found = strcmpi(inputVal, paramNames);
+                if ~any(found)
                     throw(exceptions.nameNotRecognised(sprintf('Parameter %s not recognized', inputVal)));
                 end
-                param = inputVal;
+                param = paramNames{find(found, 1)};
 
             elseif isnumeric(inputVal)
-                paramIndex = floor(inputVal);
```
DrPaulSharp, line 201: This line presumably ensure the numeric input is an integer. Are you happy that it isn't necessary?

API/projectClass/resolutionsClass.m
```diff
@@ -260,11 +260,11 @@ function displayResolutionsObject(obj, showPriors)
...
                 end
             elseif isText(param)
-                if ~strcmpi(param, paramList)
+                found = strcmpi(param, paramList);
+                if ~any(found)
                     throw(exceptions.nameNotRecognised(sprintf('Unrecognised %s name %s', parameterType, param)));
-                else
-                    thisPar = param;
                 end
+                thisPar = paramList(find(found, 1));
```
DrPaulSharp, line 267: Most of the other `paramList`-like arrays are cell arrays, sometimes coming after a call to `cellstr`. Is it worth doing that here and in the backgrounds class for consistency? And is it then possible to put this sequence into a function?

DrPaulSharp (approved):
This all looks good, I've a couple of things for you to consider in the comments. Also, could you please rebase this PR and remove mentions of oil-water calculations.

API/projectClass/layersClass.m
StephenNneji: Yes, its a bad idea as we cannot assume the users intentions, if `inputVal` is 1.5, does the user want 1 or 2? Also its not done anywhere else.

DrPaulSharp (approved):

PR #358, opened 2025-03-04
Adds docstring for parameter class and some refactors

StephenNneji:
Refactors the `projectClass` to reduces the number of lines in the file. Breaks out the parameter methods in the file into a mixin class. Removes non-keyword set method such as `setParameterValue` to unify with other properties like Bulk-in.


API/projectClass/parametersClass.m
```diff
@@ -1,26 +1,48 @@
 classdef parametersClass < tableUtilities
-    % This is the class definition for
-    % the parameters block.
-    
+    % ``parametersClass`` manages the parameters for the project. It provides methods to add, update and remove parameters.
+    % Each parameter is stored as a row in a table and consist of a name, a value with its minimum and maximum limits, a flag indicating
```
DrPaulSharp, line 3: This should be "consists" here.

API/projectClass/parametersClass.m
```diff
@@ -1,26 +1,48 @@
 classdef parametersClass < tableUtilities
-    % This is the class definition for
-    % the parameters block.
-    
+    % ``parametersClass`` manages the parameters for the project. It provides methods to add, update and remove parameters.
+    % Each parameter is stored as a row in a table and consist of a name, a value with its minimum and maximum limits, a flag indicating 
+    % if the parameter should be fitted in the calculation, and information for Bayesian calculations such as whether the prior is uniform 
+    % or gaussian, then mu and sigma for gaussian priors. ``parametersClass`` will be initialised with a default first parameter if no 
+    % arguments are provided otherwise the provided arguments will be used to create the first parameter.
```
DrPaulSharp, line 6: I'm wondering if this would read clearer as:

"if the parameter should be fitted in the calculation, and priors for Bayesian calculations. The type of the prior (either "uniform", "gaussian", or "jeffreys") is included alongside the mean (mu) and standard deviation (sigma) of the Gaussian distribution for Gaussian priors. ``parametersClass`` will be initialised . . ."

Note that Alex has filed the PR for the jeffreys prior, and it will probably be merged today.

DrPaulSharp (commented):
This looks great, just a bit of tidying up to do.

Also, is there any point in retaining the `setValues`, `setLimits`, etc methods in `parametersClass` since they're not called and we don't recommend using them?

API/projectClass/parametersClass.m
```diff
@@ -333,44 +511,100 @@ function displayTable(obj, showPriors)
...
             end
-            priors = {priorType, muValue, sigmaValue};
+            priors = {priorType, mu, sigma};
         end
 
-        function validateLimits(minLimit, value, maxLimit)
-            % Validate the value, lower and upper limit variables
-            if ~(isnumeric(minLimit) && isnumeric(value) && isnumeric(maxLimit))
-                throw(exceptions.invalidType('min, value, and max must be numbers'));
+        function validateLimits(~, min, value, max)
```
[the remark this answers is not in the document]
StephenNneji: These functions were previously static but I did not want them documented in the class as they have no direct use for the user so I made them private and since they don't use the `obj` argument I set it to the tilde. Might just be better to move them into utils instead

API/projectClass/projectParametersMixin.m
```diff
@@ -0,0 +1,547 @@
...
+                    dims = size(laysTable);
+                    for m = 1:dims(1)
+                        for n = 1:dims(2)
+                            tablePar = laysTable{m,n};   % Should be a string
+                            if isequal(findParam, tablePar)
+                                obj.layers.varTable(m,n) = {''};
+                            end
+                        end
+                    end
+                end
```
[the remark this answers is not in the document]
DrPaulSharp: For example, renaming a parameter does not trigger this code block, but would produce the same problem in that there is a layer parameter that is not defined.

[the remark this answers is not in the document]
StephenNneji: https://github.com/RascalSoftware/RAT/blob/324734592fa38bd33956881a1e397af65122f621/API/projectClass/projectClass.m#L292

The code converts the row index to a string directly, so it does not work. I will remove it but we need to add some checks maybe in our `toStruct` method so we give better error messages when the users has removed a used parameter.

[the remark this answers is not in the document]
StephenNneji: It possibly works if the parameter is removed by name with exact case

API/projectClass/parametersClass.m
```diff
@@ -1,244 +1,439 @@
...
+            % max : double, default: 0.0
+            %     The maximum value that the parameter could take when fitted, default will be equal to ``value`` if this is not set.
+            % fit : logical, default: false
+            %     Whether the parameter should be fitted in a calculation.
+            % priorType : PriorTypes, default: PriorTypes.Uniform 
+            %     For Bayesian calculations, whether the prior likelihood is assumed to be ‘uniform’ or ‘gaussian’.
+            % mu : double, default: 0.0
+            %     If the prior type is Gaussian, the mu and sigma values describing the Gaussian function for the prior likelihood.
+            % sigma : double, default: Inf
+            %     If the prior type is Gaussian, the mu and sigma values describing the Gaussian function for the prior likelihood.
```
DrPaulSharp, line 111: Use mean and standard deviation here.

API/projectClass/projectParametersMixin.m
```diff
@@ -0,0 +1,576 @@
...
+            % max : double, default: 0.0
+            %     The maximum value that the bulk-out parameter could take when fitted, default will be equal to ``value`` if this is not set.
+            % fit : logical, default: false
+            %     Whether the bulk-out parameter should be fitted in a calculation.
+            % priorType : PriorTypes, default: PriorTypes.Uniform 
+            %     For Bayesian calculations, whether the prior likelihood is assumed to be ‘uniform’ or ‘gaussian’.
+            % mu : double, default: 0.0
+            %     If the prior type is Gaussian, the mu and sigma values describing the Gaussian function for the prior likelihood.
+            % sigma : double, default: Inf
+            %     If the prior type is Gaussian, the mu and sigma values describing the Gaussian function for the prior likelihood.
```
DrPaulSharp, line 246: Use mean and standard deviation here.

API/projectClass/projectParametersMixin.m
```diff
@@ -0,0 +1,576 @@
...
+            % max : double, default: 0.0
+            %     The maximum value that the bulk-in parameter could take when fitted, default will be equal to ``value`` if this is not set.
+            % fit : logical, default: false
+            %     Whether the bulk-in parameter should be fitted in a calculation.
+            % priorType : PriorTypes, default: PriorTypes.Uniform 
+            %     For Bayesian calculations, whether the prior likelihood is assumed to be ‘uniform’ or ‘gaussian’.
+            % mu : double, default: 0.0
+            %     If the prior type is Gaussian, the mu and sigma values describing the Gaussian function for the prior likelihood.
+            % sigma : double, default: Inf
+            %     If the prior type is Gaussian, the mu and sigma values describing the Gaussian function for the prior likelihood.
```
DrPaulSharp, line 370: Use mean and standard deviation here.

API/projectClass/projectParametersMixin.m
```diff
@@ -0,0 +1,576 @@
...
+            % max : double, default: 0.0
+            %     The maximum value that the scalefactor parameter could take when fitted, default will be equal to ``value`` if this is not set.
+            % fit : logical, default: false
+            %     Whether the scalefactor parameter should be fitted in a calculation.
+            % priorType : PriorTypes, default: PriorTypes.Uniform 
+            %     For Bayesian calculations, whether the prior likelihood is assumed to be ‘uniform’ or ‘gaussian’.
+            % mu : double, default: 0.0
+            %     If the prior type is Gaussian, the mu and sigma values describing the Gaussian function for the prior likelihood.
+            % sigma : double, default: Inf
+            %     If the prior type is Gaussian, the mu and sigma values describing the Gaussian function for the prior likelihood.
```
DrPaulSharp, line 493: Use mean and standard deviation here.

DrPaulSharp (approved):
looks good, just some mu and sigma comments to update.

PR #374, opened 2025-03-21
Fixes bug for standard layers without hydration

DrPaulSharp:


alexhroom (approved):
works as expected! :)

PR #379, opened 2025-03-24
Refactors Abeles calculation

DrPaulSharp:
For a standard matlab implementation, this PR gives a factor of 3 speedup to `findkn`, resulting in a 25% speedup for the code as a whole. This is for a DREAM benchmark using standard layers and custom layers.


targetFunctions/common/reflectivityCalculations/abeles/findkn.m
```diff
@@ -1,30 +1,6 @@
-function kn = findkn(k0,sld)
+function kn = findkn(k0, sld)
```
StephenNneji, line 1: Will it be useful to keep some comments?

StephenNneji (approved):
Thanks, good Job 👍. Please see comment

targetFunctions/common/reflectivityCalculations/abeles/findkn.m
DrPaulSharp: Hmm, I was wondering about that. I'll take another look and see.

PR #380, opened 2025-03-25
Improves DREAM log-prior and adds Jeffreys prior

### LINKED ISSUE #353: implement Jeffreys prior in DREAM
[Jeffreys priors](https://en.wikipedia.org/wiki/Jeffreys_prior) are supported by NS but not DREAM - see whether they can be added to DREAM and add them if possible.

### LINKED ISSUE #370: Improve and clarify DREAM log-prior implementation
**Edit: it seems that scaledGaussPrior calculates the log-prior if a Gaussian prior is used. However, I don't think it applies log to uniform priors, and I think we should still use inverse transform sampling (or maybe even just the NS resampleParams routine?)**

-----

In the DREAM routine `scaledGaussPrior.m`:

https://github.com/RascalSoftware/RAT/blob/8c7ffa0e4c63946c9d66368b334ae90dca04dbdd/minimisers/DREAM/scaledGaussPrior.m#L26-L36

first of all I think this should use inverse transform sampling, as we've already found with NS that this way of doing scaling causes issues with the location of mu and sigma.

secondly, I'm not quite sure why there's a sign flip at the end? this makes the prior negative for a positive-valued parameter.

finally, in `calcDensity.m` we have:

https://github.com/RascalSoftware/RAT/blob/8c7ffa0e4c63946c9d66368b334ae90dca04dbdd/minimisers/DREAM/functions/calcDensity.m#L57-L63

the comment `% Does it even need to be log?` should be answered. to me unless I'm missing something the answer should be yes, but Coder crashes if you add the log because the prior values for Gaussians are negative.

alexhroom:
This PR fixes #370 and fixes #353 with refactors and improvements to the DREAM algorithm.

The changes to log-prior seem to create some minor improvements in chi-squared (optimisation benchmark goes from 0.35 to 0.31 for DREAM)

Changes:
- renamed `evaluateModel` to `calcLogLikelihood`. It turned out that the output of `evaluateModel`, `fx`, was just the transpose of `log_L` so it could be done directly.
- removed `fx` (see previous bullet point) and `Meas_info` (unused)
- replaced `calcProposal` with `calcLogPrior`, which now directly and correctly calculates log-prior for uniform and Gaussian, and also adds Jeffreys'.
- removed `scaledGaussPrior` as it is no longer used (due to the implementation of `calcLogPrior`).
- renamed ratDREAM.m to DREAM.m


minimisers/DREAM/functions/initializeDREAM.m
```diff
@@ -46,18 +42,15 @@
     [x] = boundaryHandling(x,paramInfo);
 end
 
-% Now evaluate the model ( = pdf ) and return fx
-fx = evaluateModel(x,DREAMPar,Meas_info,ratInputs);
+% Now evaluate the model ( = pdf ) and return log-likelihood
+log_L_x = calcLogLikelihood(x,DREAMPar,ratInputs);
```
DrPaulSharp, line 46: Would it be worth using `Log_L_x` as the output variable for `calcLogLikelihood` too?

minimisers/DREAM/runDREAM.m
```diff
@@ -93,8 +87,7 @@
 
 
 % Run the sampler....
-Meas_info = struct('Y',0,'N',0);
-[chain,dreamOutput,~] = ratDREAM(DREAMPar,ParInfo,Meas_info,ratInputs);
+[chain,dreamOutput,~] = ratDREAM(DREAMPar,ParInfo,ratInputs);
```
DrPaulSharp, line 90: Given the other version of this file has been deleted, could you please rename `ratDREAM` as `DREAM`.

DrPaulSharp (approved):
This looks great, thanks very much for taking such care over the issue we found previously. A better, faster (and possibly also harder and stronger) algorithm isn't a bad result!

Just a couple of quick things and a rebase required

Issue #371, opened 2025-03-21
Unhelpful error if you forget to add layers


alexhroom:
If you forget to add layers to your project, when you try to add the contrast you get the unhelpful error:

```
Error using contrastsClass/validateContrastModel (line 249)
Only one model value is allowed for custom models
```

rather than an error telling you that your layers are not defined.

To reproduce: try to run the standardLayersDSPCScript with line 44 (`problem.addLayerGroup(Layers)`) removed.

Issue #343, opened 2025-02-12
Unused properties in input and output structs


alexhroom:
- [x] `bayesResults.dreamOutput.modelOutput` is unused - it is set to zero on initialisation and is never changed. - removed in #369 
- [x] `problemStruct.otherParams` and `problemStruct.otherLimits`  are not necessary - removed in #356 
- [x] For the Priors struct , only `priors.priorNames` and `priors.priorValues` are used the remaining properties e.g. `priors.bulkIn` etc. should be removed. https://github.com/RascalSoftware/RAT/pull/348

We should consider removing placeholders from users facing code (core calculation should not be touched) to avoid confusing the users since all these things will show up in the documentation API reference. All can be re-added as needed after the release

- [x] `QzShift` - this is still present in the problemStruct which affects the Python code - removed in #356 
- [x] `problemStruct.oilChiDataPresent`, `contrastClass.oilChiData` and the oilWater projects - this show up in the docs which can be confusing  - removed in #356 
- [x]  `Magnetic` and `MagneticDomains` - these are very easy to add back but should be removed before release to avoid confusion - removed in #356 
- [x] `problemStruct.repeatLayers` not implemented - implemented in #383

alexhroom:
@DrPaulSharp should this be closed now #383 is merged?

DrPaulSharp:
Yes, it should.

PR #383, opened 2025-03-26
Implements repeat layers

### LINKED ISSUE #296: Repeat Layers is still unimplemented

DrPaulSharp:
Closes #296, #343, and #371.


DrPaulSharp:
Got my issue numbers mixed up . . .

StephenNneji (commented):
Looks good, please see comments.

API/projectClass/contrastsClass.m
```diff
@@ -234,6 +244,17 @@
             output = allowedNames{find(found, 1)};
         end
 
+        function output = validatePositiveInteger(~, input)
+            if isempty(input)
+                output = [];
+                return
+            end
+            if ~(isscalar(input) && mod(input, 1) == 0  && input > 0)
+                throw(exceptions.invalidValue('The input "%s" must be a whole number greater than zero.', input))
```
StephenNneji, line 253: Please add a test for this line, Missing `sprintf`

API/projectClass/contrastsClass.m
```diff
@@ -139,6 +141,10 @@
                 end
             end
 
+            if ~strcmpi(modelType, modelTypes.StandardLayers.value) && any(contrastRepeatLayers-1)
```
StephenNneji, line 144: I think it might be better if this warning was shown when the user sets the `repeatLayers` value then it can be reset back to 1.

DrPaulSharp: Hmm, the problem is that it doesn't know the model type when adding/setting the contrast. Do we want to change the methods such that the model type must always be supplied?

StephenNneji (approved):

PR #385, opened 2025-04-01
Refactors Abeles Calculation

DrPaulSharp:
Closes #383 This PR tidies up the Abeles calculation and makes the matrix multiplication explicit. This results in a modest performance improvement in the compiled code.


alexhroom (approved):
looks good to me, and we agree on benchmarks

PR #390, opened 2025-04-08
Introduces ability to change number of threads in tests

DrPaulSharp:


testScript.m
```diff
@@ -10,10 +10,11 @@
                                  'Producing', reportFormat, ...
                                  'IncludingSubfolders', true);
 
+numThreads = feature('numCores');
 disp('Starting Parallel Pool')
 % if parallel pool has already been started, continue with it
 if isempty(gcp('nocreate'))
-    parpool('threads');
+    parpool('threads', numThreads);
```
StephenNneji, line 17: Why not `parpool('threads', feature('numCores'));` ?

DrPaulSharp: My thinking was someone might want to set an explicit number of cores, so it would be neater for them to change the value of a variable.

StephenNneji: In that case,  best to make testScript a function 

```
function testScript(numThreads)
    arguments
        numThreads {isnumeric} = feature('numCores')
    end
```

PR #411, opened 2025-08-14
Ensures the plot event always calculates SLD

### LINKED ISSUE #408: Plot event should always show SLD plot
For simplex and DE, ensure that for the iteration where the plot event will be triggred, calcSldDuringFit should be true

Important for   https://github.com/RascalSoftware/RasCAL-2/issues/106

StephenNneji:
Closes #408

### DIFF
--- a/minimisers/simplex/fMinSearch.m
+++ b/minimisers/simplex/fMinSearch.m
@@ -63,7 +63,6 @@ tolx = optimget(options,'TolX',defaultopt,'fast');
 tolf = optimget(options,'TolFun',defaultopt,'fast');
 maxfun = optimget(options,'MaxFunEvals',defaultopt,'fast');
 maxiter = optimget(options,'MaxIter',defaultopt,'fast');
-% funValCheck = strcmp(optimget(options,'FunValCheck',defaultopt,'fast'),'on');
 
 switch dis      % Changed from TMW fminsearch
     case {'notify','notify-detailed'}
@@ -89,18 +88,6 @@ problemStruct = varargin{1};
 controls = varargin{2};
 params = varargin{3};
 doPlotEvent = hasPlotHandler();
-% if funValCheck
-    % Add a wrapper function, CHECKFUN, to check for NaN/complex values without
-    % having to change the calls that look like this:
-    % f = funfcn(x,varargin{:});
-    % x is the first argument to CHECKFUN, then the user's function,
-    % then the elements of varargin. To accomplish this we need to add the 
-    % user's function to the beginning of varargin, and change funfcn to be
-    % CHECKFUN.
-%     varargin = [{funfcn}, varargin];
-%     funfcn = @checkfun;
-% end
-% 
 n = numel(x);
 
 % Initialize parameters
@@ -199,14 +186,13 @@ for j = 1:n
         y(j) = zero_term_delta;
     end
     v(:,j+1) = y;
-    if doPlotEvent
+    if doPlotEvent && j==n 
         controls.calcSLD = true;
     end
     x(:) = y; [f, result] = funfcn(x, problemStruct, controls, params);
-    controls.calcSLD = false;
     fv(1,j+1) = f;
 end
-
+controls.calcSLD = false;
 % sort so v(1,:) has the lowest function value
 [fv,j] = sort(fv);
 v = v(:,j);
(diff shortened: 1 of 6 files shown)


minimisers/simplex/fMinSearch.m
```diff
@@ -63,7 +63,7 @@
 tolf = optimget(options,'TolFun',defaultopt,'fast');
 maxfun = optimget(options,'MaxFunEvals',defaultopt,'fast');
 maxiter = optimget(options,'MaxIter',defaultopt,'fast');
-funValCheck = strcmp(optimget(options,'FunValCheck',defaultopt,'fast'),'on');
+% funValCheck = strcmp(optimget(options,'FunValCheck',defaultopt,'fast'),'on');
```
DrPaulSharp, line 66: Should this line be removed?

minimisers/simplex/fMinSearch.m
```diff
@@ -85,19 +85,21 @@
 % Convert to function handle as needed.
 % funfcn = fcnchk(funfcn,length(varargin));
 % Add a wrapper function to check for Inf/NaN/complex values
-controls = varargin{2};
 problemStruct = varargin{1};
-if funValCheck
+controls = varargin{2};
+params = varargin{3};
+doPlotEvent = hasPlotHandler();
+% if funValCheck
```
DrPaulSharp, line 92: Should this line be removed?

minimisers/simplex/fMinSearch.m
```diff
@@ -85,19 +85,21 @@
...
     % x is the first argument to CHECKFUN, then the user's function,
     % then the elements of varargin. To accomplish this we need to add the 
     % user's function to the beginning of varargin, and change funfcn to be
     % CHECKFUN.
-    varargin = [{funfcn}, varargin];
-    funfcn = @checkfun;
-end
+%     varargin = [{funfcn}, varargin];
+%     funfcn = @checkfun;
+% end
```
DrPaulSharp, line 102: Should these lines be removed?

minimisers/simplex/fMinSearch.m
```diff
@@ -191,7 +199,11 @@
         y(j) = zero_term_delta;
     end
     v(:,j+1) = y;
-    x(:) = y; [f, result] = funfcn(x,varargin{:});
+    if doPlotEvent
+        controls.calcSLD = true;
+    end
+    x(:) = y; [f, result] = funfcn(x, problemStruct, controls, params);
+    controls.calcSLD = false;
```
DrPaulSharp, line 206: Can the value of `doPlotEvent` change within this loop? If not, it would make sense to move the `controls.calcSLD` lines outside of the loop rather than continually turning on and off.

DrPaulSharp: Also, we don't actually do the plot until after the loop, so presumerably we only need to calculate the SLD on the last iteration of the loop. I'm very much not a fan of `if j == n` but . . . It's one to think about, we want to avoid all unnecessary SLD calculations.

minimisers/simplex/fMinSearch.m
```diff
@@ -257,13 +272,15 @@
...
+    x(:) = xr; [fxr, result] = funfcn(x, problemStruct, controls, params);
     func_evals = func_evals+1;
     
     if fxr < fv(:,1)
         % Calculate the expansion point
         xe = (1 + rho*chi)*xbar - rho*chi*v(:,end);
-        x(:) = xe; [fxe, result] = funfcn(x,varargin{:});
+
+        x(:) = xe; [fxe, result] = funfcn(x, problemStruct, controls, params);
+        controls.calcSLD = false;
```
DrPaulSharp, line 283: Is this in the right place? There are other evaluations of `funfcn` in the `else` clause of the `if` statement on line 278 that are not followed by `controls.calcSLD = false;`

minimisers/simplex/fMinSearch.m
```diff
@@ -334,7 +351,8 @@
 %         fprintf('%s \n', 'func_evals = ')
 %         fprintf('%s \n', num2str(func_evals))
     end
-    if rem(itercount, controls.updatePlotFreq) == 0   
+    controls.calcSLD = false;
+    if doPlotEvent && rem(itercount, controls.updatePlotFreq) == 0
```
DrPaulSharp, line 355: Again this implies that we only need the final evaluation from the while loop, but I can see that that may be very difficult to determine.

DrPaulSharp (commented):
Looks good overall, just a few things to think about in the simplex algorithm.

minimisers/simplex/fMinSearch.m
StephenNneji: I will remove as this is already handled by line 353. I probably added this when I hoped to decipher the else branch

minimisers/simplex/fMinSearch.m
StephenNneji: Yes, for now I will leave as is except if optimization is needed

DrPaulSharp (approved):

PR #426, opened 2025-10-06
Fixes live plot issue

### LINKED ISSUE #425: useLivePlot is not updating
When I enable live plot, the actual figure doesn't seem to update until the run has finished.

StephenNneji:
Fixes #425


DrPaulSharp (approved):
This change fixes the problem.

One thing I have noticed though, is that the lines change colour on the final plot when the simplex/de run is finished. Any idea why that is?

PR #428, opened 2025-11-13
Revamp the build system

StephenNneji:
Redesign build system to use a single self-hosted server. GitHub runners are then used to build mex and deploy.
Also adds build for arm64 Macs


DrPaulSharp (approved):
Looks good, and the build works well.

PR #431, opened 2026-02-03
Added utilities for saving and loading RasCAL2 files....

arwelHughes:


utilities/misc/loadR2.m
```diff
@@ -0,0 +1,15 @@
+
+function [problem,varargout] = loadR2(dirName)
+% Loads in a RasCAL2 project...
+
+% Load the project...
+projectFile = fullfile(dirName,'project.json');
+problem = jsonToProject(projectFile);
+
+% If there is a second argument, also load results..
+if nargout > 1
```
StephenNneji: @arwelHughes  This does not convert _controls.json_ is this intentional?

tests/testJsonToProject/testJsonToProject.m
```diff
@@ -54,13 +63,9 @@ function testJsonResultConversion(testCase, resultFile)
             end
```
abuts, line 64: request:
This may be previous contribution, but this test tests nothing. Structure contains no properties. 
row 60 should be:
```
            props = fieldnames(result);
```
And tests are failing if you do this. Seems on shape only but...

tests/testJsonToProject/testJsonToProject.m
```diff
@@ -69,6 +74,54 @@ function testJsonControlsConversion(testCase)
...
+            testCase.verifyEqual(exist(fullfile(pwd, outFolder, 'controls.json'), 'file'), 2)
+            for i=1:project.customFile.rowCount
+                customFile = project.customFile.varTable{i, 2};
+                testCase.verifyEqual(exist(fullfile(pwd, outFolder, customFile), 'file'), 2)
+            end
+
+            [project2, result2] = loadR2(outFolder);
+            props = properties(project);
+            for i = 1:length(props)
+                % verifies the problem name, model type and geometry
```
abuts: Request if I understand it correctly:
I am puzzled. Why this loop? Nothing inside it depends on `i`  or `props` -- you just run the same check 16 times. Should loop be removed?

All `prop` check is correctly perfromed at rows 120-124. 

Test between `for` and `end`  partially repeats `120-124` but ok, this way of testing may be more beneficial or clear, but why to run it in a loop?

tests/testJsonToProject/testJsonToProject.m
```diff
@@ -69,6 +74,54 @@ function testJsonControlsConversion(testCase)
...
+       function testR2Conversion(testCase, r2Tests)
+            scriptName = r2Tests{3};
+            outFolder = ['save_' scriptName];
+           
+            copyfile(fullfile(getappdata(0, 'root'), 'examples', r2Tests{1}, r2Tests{2}), ...
+                     fullfile(pwd))
+            evalc(scriptName);
+            
+            controls = controlsClass();
+            [~, project, result] =  evalc('RAT(problem, controls);');
```
abuts: Suggestion:
I can vageuly understand and accept why one would use `evalc` in 85, though IMHO function handle would be much better. But in 88! Why?
Why to use:
```
[~, project, result] =  evalc('RAT(problem, controls);');
```
instead of  just:
```
 [project, result] =  RAT(problem, controls);
```
It works faster and much more clear code.

abuts (changes requested):
It not to for me to make suggestions on code quality at this stage of a project, but IMHO, `cd` within a function is a very bad practice. It is usually ok to do `cd` on top level tests ensuring `onCleanup` returns to intial folder in any case, but `cd` within a routine usually causes bunch of maintenance problem on a long run. `to/from json` insides look also a bit crude.

Code works and does the job. There are 2 bugs and 1 suggestion within the test suite itself.

tests/testJsonToProject/testJsonToProject.m
StephenNneji: I have removed now, I refactored and forgot to remove loop

tests/testJsonToProject/testJsonToProject.m
StephenNneji: I was trying to hide output, but I can do the same with controls.display = 'off' so I will replace

tests/testJsonToProject/testJsonToProject.m
StephenNneji: I have removed

StephenNneji:
> It not to for me to make suggestions on code quality at this stage of a project, but IMHO, `cd` within a function is a very bad practice. It is usually ok to do `cd` on top level tests ensuring `onCleanup` returns to intial folder in any case, but `cd` within a routine usually causes bunch of maintenance problem on a long run. `to/from json` insides look also a bit crude.
> 
> Code works and does the job. There are 2 bugs and 1 suggestion within the test suite itself.

I have refactored the saveR2 function to remove the cd. I also fixed typo in standardLayersDSPCSheet.mlx

abuts (approved):
Looks good

PR #435, opened 2026-03-23
Fixes bug in contrast shift value and starts shift value at 0

DrPaulSharp:


utilities/plotting/plotRefSLDHelper.m
```diff
@@ -72,24 +73,24 @@ function plotRefSLDHelper(data, noDelay, linearX, q4, showErrorBar, showGrid, sh
...
-    end   
+    end
+
     hold on
     xlabel('$\textrm{Q}_{z} (\AA^{-1})$', 'Interpreter', 'Latex') 
     ylabel('Reflectivity', 'Interpreter', 'Latex') 
     lines = cell(numberOfContrasts, 1);
-    mult = 1;
-    q4Data = 1;
+    q4Data = 1.0;
```
StephenNneji, line 82: As all numbers in MATLAB are double not sure if this makes a difference

StephenNneji (approved):
Thanks, works as advertised

utilities/plotting/plotRefSLDHelper.m
DrPaulSharp: Old habits dying hard I think.

PR #440, opened 2026-05-22
Modified SLD profiles.

arwelHughes:
This PR fixes two issues in RasCAL2:

- Closes Issue RascalSoftware/RasCAL-2#215: This bug was caused by the total range for the SLD plots not being sufficient when large roughness' are used in the last layer if standard layer models. 'makeSLDProfile' has been modified to extend the range of the profile to cover the 99% interval of the outer error function.

- Closes Issue RascalSoftware/RasCAL-2#205: SLD profiles should always be aligned to the substrate edge. This happens automatically for substrate/liquid profiles, but for air/substrate samples should be aligned to the other end. To fix this, 'plotRefSLD' has been modified to align the substrates in A/L projects by padding the start of the SLD's.


buildScript.m
```diff
@@ -16,7 +16,7 @@
 
 compilePath = fullfile(root,'compile','customWrapper');
 cd(compilePath);
-wrapperMexBuild;
+%wrapperMexBuild;
```
DrPaulSharp, line 19: The build for this pull request is failing, this change is probably the reason why.

targetFunctions/common/makeSLDProfiles/makeSLDProfile.m
```diff
@@ -1,54 +1,76 @@
...
-            thisBox = asymconvstep(x,layerThickness,thisBoxCentre,layerRoughness,nextLayerRoughness,layerSLD);
-            Lays(:,i+(numberOfLayers*(n-1))) = thisBox;
-            lastBoxEdge = thisBoxCentre + (0.5 * layerThickness);
+    outerLayerCentre = totalThickness - (layers(end,1)/2);
+
+    % Find the point which covers 99% of the outer error function..
+    outerLayerTailLim = (erfcinv(0.01) * sqrt(2) * layers(end,3)) + outerLayerCentre;
+
+    % We need to make sure the total SLD range includes this...
+    outerLayerTailExtension = outerLayerTailLim - outerLayerCentre;
```
DrPaulSharp, line 16: The variables `outerLayerCentre` and `outerLayerTailLim` is only used here, so we should calculate `outerLayerTailExtension` directly by replacing lines 10, 13 and 16 with:

```
outerLayerTailExtension = (erfcinv(0.01) * sqrt(2) * layers(end,3));
```

DrPaulSharp (commented):
One point here, assuming all is well, I'll work to merging this work through #443.

PR #424, opened 2025-09-29
Fixes empty layers bug

DrPaulSharp:
Fixes #419 

This PR fixes the bug that occurs where a contract containing no layers is viewed as a single layer of zero thickness. This was introduced in the routine `groupLayersMod`, where it was necessary to ensure `outputLayers` was defined on all branches of the if statement. By replacing the line of zeros with an appropriately sized empty array, we retain all code functionality, ensuring the following routines, `applyHydration` and `makeSLDProfile`, behave as expected when there are no layers. I have also confirmed that the plot looks as we expect with this change.

@arwelHughes this PR means there is no need to add the check to `makeSLDProfile` that we discussed this morning.

PR #442, opened 2026-06-08
Enables empty contrast models for domains calculations

DrPaulSharp:
Following from #424, the PR enables RAT to run with contrasts containing empty models for domains calculations. As before this calculation is then the interface between two bulk phases. By definition, the domains have no effect on the calculation, since there are no domain-containing layers between the bulk phases. Hence, the two domain curves will always be identical in this case.

I have confirmed on testing that the results of a zero-layer calculation for the `normal` and `domains` calculation types are the same.


StephenNneji (approved):
Thanks for fixing this, works as advertised.

PR #444, opened 2026-07-06
Refactors NS and Dream to exit quicker when stopped and refactors alignALProfiles so it works for bayes and live plots.

StephenNneji:
- These algorithms should now return an empty result when stopped and return quicker than before
- Refactors `alignALProfiles` and makes it work for bayes and live plots.

### DIFF
--- a/utilities/plotting/alignALProfiles.m
+++ b/utilities/plotting/alignALProfiles.m
@@ -34,11 +34,9 @@ lengths = zeros(size(sldProfiles));
 for i = 1:numel(sldProfiles)
     lengths(i) = size(sldProfiles{i}, 1);
 end
-% Get max length and its index
-[maxLen, maxPos] = max(lengths);
-maxPos = maxPos(1);
-maxLen  = maxLen (1);
 
+% Get max length and its index
+[maxLen, maxPos] = max(lengths, [], "all");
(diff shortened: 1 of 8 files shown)


utilities/plotting/alignALProfiles.m
```diff
@@ -0,0 +1,71 @@
...
+
+% Find the length of the longest profile.
+lengths = zeros(size(sldProfiles));
+for i = 1:numel(sldProfiles)
+    lengths(i) = size(sldProfiles{i}, 1);
+end
+% Get max length and its index
+[maxLen, maxPos] = max(lengths);
+maxPos = maxPos(1);
+maxLen  = maxLen (1);
```
DrPaulSharp: Are these lines necessary given the docs say the index returned is "the first occurrence of the maximum value of A"?
https://uk.mathworks.com/help/matlab/ref/double.max.html#d127e1109685

DrPaulSharp (approved):
All looks good, works for all plot types and noticeable faster exit on interruption. Just one point to consider.

utilities/plotting/alignALProfiles.m
StephenNneji: I had to do `[maxLen, maxPos] = max(lengths, [], "all");` otherwise coder build fails

PR #446, opened 2026-07-09
The full bulk_in array is now passed into custom function

StephenNneji:
Fix #445 

Event mods to facilitate running in python


DrPaulSharp (approved):
All looks good. It occurs to me that this represents a change in behaviour from RasCAL 1, so we'll need to bear that in mind if anyone spots a problem with their custom functions.

